PR #14, opened 2024-09-26
Make sure dp-creator-ii can start with no args

### LINKED ISSUE #12: Should start without any command line arguments
Plain `dp-creator-ii` currently fails because it is trying to read `dp_creator_ii/app/config.json` which isn't written if there are no CLI arguments.

mccalluc:
- Fix #12


mccalluc:
Confirming that on main I can run `dp-creator-ii` with no arguments with no problems.

PR #16, opened 2024-09-26
Modularize UI

### LINKED ISSUE #5: Modularize UI code
At the very least, each tab should get a separate python file. There are [Shiny modules](https://shiny.posit.co/py/docs/modules.html). Maybe it would also make sense to separate the UI and server code?

mccalluc:
- Fix #5

We don't actually need [shiny modules](https://shiny.posit.co/py/docs/modules.html#how-to-use-modules): They will be useful when we need to reuse UI elements, but each tab is only used once, so regular functions suffice.


ekraffmiller (approved):
looks good!

PR #28, opened 2024-09-30
 unit code snip

### LINKED ISSUE #20: On CI, just download one browser?
It doesn't add much to the runtime of the tests, but if we're testing on one browser, can we just download one browser on CI? Currently grabbing webkit and firefox and chromium.

mccalluc:
- Fix #20
- Incorporates #16 and #17: They should be reviewed and merged first.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,7 +4,7 @@ source = .
 
 omit =
   # TODO
-  app.py
+  dp_creator_ii/app/*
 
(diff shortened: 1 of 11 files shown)


mccalluc:
I want to merge in `12-start-with-no-args`, but the merge history is already too complicated here. Wait until Ellen has had a chance to the other PRs, then pick this up again.

mccalluc:
Redo this, starting from main.

PR #35, opened 2024-10-02
Create mock data and plot

### LINKED ISSUE #27: Whisker plot, with a dashed line for delta
Don't even worry about connecting it to any particular data, but just figure out which charting library to use to so we can have:
- whiskers on the bars
- dashed horizontal line across the full width
- perhaps grey-out the bars below that line
- no labels on vertical axis. (With mock data, the absolute number of counts wouldn't matter, so we don't want to label the axis.)

### LINKED ISSUE #30: Given column names and bounds, create dataframe of normally distributed values
We need fake data to demonstrate the effects of DP parameter choices. My sense is that rather than a random data set, this should be deterministic, and the only noise should come from the DP parameters. (ie: if you could set epsilon to zero, and gave yourself a lot of bins, it should look like a normal distribution w/o noise.)

mccalluc:
- Fix #30
- Fix #27 

Down the road, we wouldn't use this plot with this data -- instead it would be binned and the bin counts would be privatized and that's what we plot -- but for now the two features complement each other.

Haven't made a test for the plot function. There are tools that can do image comparisons to make sure output is stable, but I'm not sure we need anything that heavy weight at this point. Another approach would be to mock matplotlib and make sure the appropriate methods are called.


dp_creator_ii/app/analysis_panel.py
```diff
@@ -1,16 +1,34 @@
...
+    @render.plot()
+    def plot_preview():
+        min_x = 0
+        max_x = 100
+        df = mock_data({"col_0_100": ColumnDef(min_x, max_x)}, row_count=20)
+        return plot_error_bars_with_cutoff(
+            df["col_0_100"].to_list(),
+            x_min_label=min_x,
+            x_max_label=max_x,
+            y_cutoff=30,
```
ekraffmiller, line 28: what is the purpose of the cutoff bar?

ekraffmiller (commented):
looks good, just a questions about the plot_preview()

dp_creator_ii/app/analysis_panel.py
mccalluc: I think that I heard from Mike that approximate (delta-epsilon) DP would censor values below a certain level, but I'm not sure exactly how that works. I ask the same question in:
- #48

ekraffmiller (approved):
question addressed, approving

PR #41, opened 2024-10-03
user provided contributions flows all the way to the generated code

### LINKED ISSUE #26: Unit of privacy code snip?
Instead of offering a download at the end, we want to show little code snippets along the way. An easy one to start with is the unit of privacy on the first tab.

mccalluc:
- Fix #26 
- New template to just define the privacy unit
- Renamings everywhere to distinguish `contributions`, a number, from `privacy_unit`, an particular OpenDP object.
- Clean up the arg parsing (so playwright will emulate a run with no command line args, but defaults supplied by argparse will be filled in)
- Introduce `ui_helpers.py` for UI code that should be reused. (For now, putting the code sample under a `<details>` tag is easy, but if we want to change it in the future, we should do it in one place.)


dp_creator_ii/app/dataset_panel.py
```diff
@@ -3,39 +3,48 @@
...
+        # We are running a Playwright test,
+        # and ARGV is polluted, so override:
+        return arg_parser.parse_args([])
+    else:
+        # Normal parsing:
+        return arg_parser.parse_args()
 
 
 def dataset_ui():
+    args = get_args()
```
ekraffmiller, line 22: Is it possible to move the command line argument processing out of the dataset_panel, and add the parsed arguments (csv_path, contributions) as parameters to dataset_ui()? That way dataset_panel.py only contains logic related to the UI Panel.

dp_creator_ii/tests/test_template.py
ekraffmiller, line 1: The changes in this file for the PR look good, but I wanted to ask about test_fill_values() - can there be more assertions here, for the other lines in the context_block?  Also in test_make_script(), the test is checking that the python script runs without error, but it doesn't check the contents of the script.  Even though you are testing the other functions individually, it could be good to test that the final script has the correct contents.

ekraffmiller (commented):
This looks good! I just have a couple of questions, see below.

dp_creator_ii/app/dataset_panel.py
mccalluc: There is still a call to `get_args` in `dataset_server` below, so I'm not sure that it would be much cleaner just to move one and not the other. I do wonder if there is some way the contributions input could get its default value from the server, but the default value isn't really something that is reacting to user input, so I don't see how that would work.

The right way to do this might be more clear to me with more familiarity with Shiny: Feel free to file a follow-up issue from this comment, if you want to be sure that we don't lose track.

dp_creator_ii/tests/test_template.py
mccalluc: For `test_fill_values`, I was thinking of it as a unit test of the template filling, so I think the best resolution is to make it simpler, so it only does that.

I will make `test_make_script` stronger, since it is more of an integration test. Thanks!

mccalluc:
Oh: applying black to the script fixture made it diverge from what we actually output. fixing...

mccalluc:
There is still work to do in this area, so I've filed a couple follow-up issues, but I think this much makes sense for now.

dp_creator_ii/app/dataset_panel.py
ekraffmiller: #51 Another issue created to analyze the possibility of moving the logic outside the component,

ekraffmiller (approved):
Comments are addressed, approving

PR #45, opened 2024-10-04
Read CSV columns from uploaded file

### LINKED ISSUE #38: Parse uploaded CSV and return column headers
- Prereq for #33

mccalluc:
- Fix #38
- More CLI input validation, to save trouble downstream
- For now, not worrying about the UI error if no file is provided: Down the road, the columns aren't even used on the first tab, but it makes debugging easier to have everything in one place temporarily
- Drop the CWD when starting the app, so relative paths supplied on the CLI resolve the same way inside the app
- Unit tests of CSV parsing, input validation
- Add file upload to end-to-end test

The pairs of `reactive.calc` and `render.text` functions seem redundant, but I ran into trouble when I tried to apply both decorators to one function. When I feel more confident about how I'm using the library may ask about this in one of the forums.


dp_creator_ii/__init__.py
```diff
@@ -1,21 +1,29 @@
...
 
 
 __version__ = "0.0.1"
 
 
+def existing_csv(arg):
+    path = Path(arg)
+    if not path.exists():
+        raise ArgumentTypeError(f"No such file: {arg}")
+    if path.suffix != ".csv":
```
ekraffmiller, line 16: Is there another way to test for type == csv?  One that examines the content of the file, rather than looking for the file extension?

dp_creator_ii/__init__.py
```diff
@@ -29,13 +37,12 @@ def get_arg_parser():
 
 def main():  # pragma: no cover
     # We call parse_args() again inside the app.
-    # We only call it here so "--help" is handled.
+    # We only call it here so "--help" is handled,
+    # and to validate inputs.
     get_arg_parser().parse_args()
```
ekraffmiller, line 42: Can we also have a test that the required arguments were passed?  When I run the app with no args, it shows a stack trace of an unhandled error. I read your comment about skipping this for now, but I'm not sure I understand how that helps with debugging?

mccalluc:
@ekraffmiller , added assertions in the end-to-end test to make sure there are no shiny errors, and fixed the error we were seeing my just returning `None` if no file has been selected. Ready for re-review.

ekraffmiller (commented):
Comments addressed, approving

PR #52, opened 2024-10-08
Remove shinywidgets

### LINKED ISSUE #43: For now, remove shinywidgets dependency
Not necessary for plain plots.

mccalluc:
- fix #43

I had introduced this dependency copy-and-pasting from tutorials, but we don't need it, and may never need it. It allows interactive widgets designed for Jupiter to work in Shiny.

Even if they come for free, I'm inclined to keep interactive visualizations out of this project. Producing PDFs at some point isn't unlikely, and dealing with interactive visualizations would be a distraction.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,6 @@ classifiers = ["License :: OSI Approved :: MIT License"]
 dynamic = ["version", "description"]
 dependencies = [
     "shiny",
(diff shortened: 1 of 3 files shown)


ekraffmiller (approved):
Looks good, there is a question about if/how we want to manage dependencies for installation, a separate issue for that is here #60

Issue #51, opened 2024-10-08
Move `get_args` out of `dataset_ui`


mccalluc:
> There is still a call to `get_args` in `dataset_server` below, so I'm not sure that it would be much cleaner just to move one and not the other. I do wonder if there is some way the contributions input could get its default value from the server, but the default value isn't really something that is reacting to user input, so I don't see how that would work.
>
> The right way to do this might be more clear to me with more familiarity with Shiny.

_Originally posted by @mccalluc in https://github.com/opendp/dp-creator-ii/pull/41#discussion_r1792184087_

PR #61, opened 2024-10-16
Add `--demo` CLI option

### LINKED ISSUE #7: Implement `--demo` option
People might not actually have a suitable CSV on hand. We should bake in a small one, so folks can install the software and immediately do something.

mccalluc:
- extract fake csv creation from #48: Once this is merged to main, I would plan to merge main to that branch, and remove the code from the notebook that is now covered here.
- Fix #7 (though we don't do anything with the data yet)
- enable doctests
- Consolidate arg parsing and demo dataset creation in one file
- Related to #51, but don't think it really fixes it.

There are a few `pragma: no cover` notes scattered over this where we are missing test coverage. If you want to hold this to a higher standard, I wouldn't object, but I think this is sufficient for now.

### DIFF
--- a/.pytest.ini
+++ b/.pytest.ini
@@ -6,5 +6,5 @@ filterwarnings =
 
 addopts = --doctest-glob '*.md' --doctest-modules --ignore dp_creator_ii/templates/ --ignore dp_creator_ii/tests/fixtures/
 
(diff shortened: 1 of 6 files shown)


ekraffmiller (approved):
Issue created to add help text about demo data #75

PR #64, opened 2024-10-16
For October 22

### LINKED ISSUE #63: Clean up "Select Dataset" after big merge

mccalluc:
Heard that the really granular PRs were hard to review, so I'll try to just put the next batch of work in one, if it seems related.
 
Original description:

> - Fix #63
>
> Move calculated fields to the next page in the flow, since we don't really need them here.
>
> File name doesn't seem to be preserved when the uploaded file is stored locally: After upload, I'm seeing the name as `0.csv`. For now, let's just not worry about echoing that back to the user, since we will be showing them their column headers.


dp_creator_ii/tests/test_app.py
```diff
@@ -36,13 +36,13 @@ def expect_no_error():
...
     csv_path = Path(__file__).parent / "fixtures" / "fake.csv"
     page.get_by_label("Choose CSV file").set_input_files(csv_path.resolve())
-    expect_visible("student_id")
     expect_no_error()
 
     page.get_by_role("button", name="Define analysis").click()
     expect_not_visible(pick_dataset_text)
     expect_visible(perform_analysis_text)
     expect_not_visible(download_results_text)
+    expect_visible("student_id")
```
ekraffmiller, line 45: the test could be updated to check for all columns in the data file

mccalluc: done!

ekraffmiller (approved):
Reviewed with Chuck, looks good

PR #67, opened 2024-10-16
rebuild requirements-dev.txt

mccalluc:
- Fix https://github.com/opendp/dp-creator-ii/security/dependabot/1

More changes than are strictly necessary. Latter on, we'd probably be more careful about an update across the board, but right now it is low risk.


ekraffmiller (approved):
looks good

PR #69, opened 2024-10-17
Text and CSV output proposals

mccalluc:
Need to consider earlier feedback from Michael on
- #48
  - [Text](https://github.com/opendp/dp-creator-ii/pull/48#discussion_r1793646657)
  - [CSV](https://github.com/opendp/dp-creator-ii/pull/48#discussion_r1793647349)

but I wanted to separate this from the main notebook so the blockers on each can be more clear.

I think I also need to become more familiar with opendp-logger before using it here.


mccalluc:
Text and CSV output has been implemented. This was useful in setting direction, but is no longer needed.

PR #87, opened 2024-10-23
Column config UI, and template fixes

### LINKED ISSUE #33: Prompt for each numeric column
Maybe it starts out as a table, but when a column is configured, the preview chart is shown below it... though this may be collapsible?

For each numeric column, we will prompt for:
- min and max (these are used both for clipping and to generate fake data)
- bins
- accuracy
  - for now, let's try just three options: "less accurate"/"default"/"more accurate", with weights of 1/2/4 respectively.
  - down the road might make this more numeric
- checkboxes for
  - mean
  - stddev
  - median? Requires candidates, which we could could take from the histogram bins...

mccalluc:
- Finish fix #33
- Make the template filling code more robust: 
  - Check for word boundaries when filling slots
  - Remember the slots that need filling, and only warn if those are unfilled at the end (because the user supplied strings may be all caps)
- More code examples in the UI, where they make sense
  -  Code samples now have titles


ekraffmiller (approved):
approved

PR #101, opened 2024-10-25
Status flow chart

mccalluc:
I'm guessing about how some of these transitions work, so if you have something else in mind, please suggest changes.


cmbz:
Looks good to me. Thank you @mccalluc

cmbz (approved):
Looks good. Thanks!

PR #102, opened 2024-10-29
Improve handling of different encodings

### LINKED ISSUE #79: Add `encoding="utf-8-sig"` when reading CSV
Files without BOM should still read cleanly, and file BOM won't have xFEFF as their first character.
```
>>> with open("dp_creator_ii/tests/fixtures/fake.csv", newline="", encoding="utf-8-sig") as csv_handle:
...     reader = csv.DictReader(csv_handle)
...     print(reader.fieldnames)
...
['student_id', 'class_year', 'assignment_type', 'grade']
```

mccalluc:
This improves our handling of different encodings, but it also involves using polars to read data when we just need the column headers.

- Fix #79

### DIFF
--- a/dp_creator_ii/utils/csv_helper.py
+++ b/dp_creator_ii/utils/csv_helper.py
@@ -1,7 +1,12 @@
-import csv
+import polars as pl
 
 
 def read_field_names(csv_path):
-    with open(csv_path, newline="") as csv_handle:
-        reader = csv.DictReader(csv_handle)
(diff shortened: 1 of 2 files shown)


Shoeboxam (approved):
Would be good to have these checks in OpenDP, and when the lib processes a lazyframe with a scan_csv, it could change the plan to ignore-errors. This way the function errors will be data-independent.

Issue #32, opened 2024-10-02
After release, controls on previous pages should be locked


mccalluc:
After talking with Mike, we decided that after making a release
- (see #31)

the inputs on the earlier pages should be locked, since for now we are not tracking the cumulative epsilon spend.

(The user can certainly just close the application and restart, since for the MVP there is now memory between sessions, but the design should discourage multiple releases.)

PR #103, opened 2024-10-30
Better handing for `--demo`

### LINKED ISSUE #75: More explanation in UI when running with `--demo`

### LINKED ISSUE #81: Handle `--demo` nicely on first tab
Either the name of the demo csv should show up in the upload widget, or we should just show the name of the demo csv in place of the widget. (Right now it is just empty.)

mccalluc:
- Fix #81: We can't actually set the value of the upload widget, but we can set placeholder text. 
- Fix #75: Add tool tips on the first tab if we're running with `--demo`

Does not address:
- #32

Eventually, we want the inputs on this screen to be frozen if they have been set from the CLI.


ekraffmiller:
Looks good! UI Suggestion:  When running with --demo, don't show the FileUpload component, since demo.csv is being used.  Include some text about demo being used instead.

ekraffmiller (approved):
Approved with a follow up suggestion for UI

mccalluc:
- Filed #113

mccalluc:
App test fails:
```
        # -- Download results --
        page.get_by_role("button", name="Download results").click()
        expect_not_visible(pick_dataset_text)
>       expect_not_visible(perform_analysis_text)
```
Looking at the trace, it looks like it might have gotten bogged down in recalculating the graphs and didn't respond to the mouseclick to move to the next page quickly enough. Is there a risk of the application being unresponsive?

For now, just try to run run the tests.

mccalluc:
- Filed #116

PR #106, opened 2024-10-30
pytest -vv works now

### LINKED ISSUE #104: `pytest -vv` -> `unexpected SystemExit`
There are tests that look at `sys.argv` and they get confused when there are pytest options that we haven't explicitly handled. Mocking should be able to fix this.

mccalluc:
- Fix #104


ekraffmiller (approved):
reviewed with Chuck, approved

PR #107, opened 2024-10-31
Only install Chromium

### LINKED ISSUE #20: On CI, just download one browser?

mccalluc:
- Fix #20
- CI download time might not be the most important thing to optimize: We could just drop this and close the issue. Thoughts?
- Another option, and probably not necessary for us, would be to configure a matrix and confirm that different browsers work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,7 +38,9 @@ jobs:
         run: pip install -r requirements-dev.txt
 
       - name: Install browsers
-        run: playwright install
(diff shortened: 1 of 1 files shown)


ekraffmiller:
I think Chrome has a much bigger user base, so maybe it is worth the extra download time, if we are only testing against one browser.

ekraffmiller (commented):
see below

ekraffmiller (approved):
looks good, approved

PR #114, opened 2024-10-31
.venv/bin/dp-creator-ii

pdurbin:
Maybe this is just me, maybe there's something wrong with my Python environment, but `dp-creator-ii` isn't in my $PATH. I had to run it as `.venv/bin/dp-creator-ii`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ $ playwright install
 Now install the application itself and run it:
 ```shell
 $ flit install --symlink
-$ dp-creator-ii
+$ .venv/bin/dp-creator-ii
 ```
 Your browser should open and connect you to the application.
 


mccalluc:
Huh: Did you run `source .venv/bin/activate`? And did it update your PATH to include your `dp-creator-ii/.venv/bin`? You shouldn't need the full path if PATH is updated.

mccalluc (commented):
Can you confirm that `activate` worked for you? The path here shouldn't be necessary.

mccalluc:
Let's keep an eye on this: If other folks hit it, something may be up. I'd be reluctant to specify the path to the virtual env, because someone might have another naming convention and then it wouldn't work for them.

PR #117, opened 2024-11-01
rename tests to match the files they test

### LINKED ISSUE #112: Make test file names mirror source code

mccalluc:
- Fix #112


ekraffmiller (changes requested):
This looks good, I just have one more suggestion. Can we have a directory structure under tests that mirrors the source code?  For example have a `tests/utils `and a `tests/utils/templates` for the current test files.  It seems trivial now, but it will make things easier to follow as the code base grows.

mccalluc:
Sure!

mccalluc:
Moved some tests down to a `utils` directory. Since
```
.../templates.py
```
is imported the same way as
```
.../templates/__init__py
```
so I think it it best to leave its tests in the util directory: `test__init__.py` would be strange.

ekraffmiller (approved):

Issue #119, opened 2024-11-01
Do mocks not work inside apps?


mccalluc:
Working on
- #105

and having some trouble: I could be misusing mocks, but I'm wondering if the problem is that the app isn't code we run directly from the pytest process: Instead it's referenced from `create_app_fixture`.

To close:
- Make a minimal reproducer
- Add a note in lessons-learned: #118

Issue #113, opened 2024-10-31
When running --demo, don't show the FileUpload component


mccalluc:
> Looks good! UI Suggestion:  When running with --demo, don't show the FileUpload component, since demo.csv is being used.  Include some text about demo being used instead.

_Originally posted by @ekraffmiller in https://github.com/opendp/dp-creator-ii/issues/103#issuecomment-2450528997_

mccalluc:
I'm happy with the status quo:
- `demo.csv` is shown in the UI, albeit as hit text rather than the actual value.
- Even if someone starts in `--demo` mode, I don't know that we need to prevent them choosing a file: It's ok to change your mind.
- Reluctant to add code that would in some sense decrease functionality.

Could reopen with user feedback, but close for now.

PR #120, opened 2024-11-01
Playwright test for `--demo`

### LINKED ISSUE #105: Playwright test for `--demo`
With more special features when running with `--demo`
- see #103

.. there should be a separate playwright test that exercises it.
- The main app test should confirm that the demo tooltips _don't_ show up.
- the demo app test should
  - make sure the tooltips _do_ show up.
  - make sure the placeholder and contributions are filled in
  - make sure they are not editable (when that is implemented)

mccalluc:
- Fix #105

Initially tried to use mocking for this, but couldn't get it to work.
- Filed #119 

... but in the end, I think this is a pretty clear approach.

When we handle this:
- #113

expand the test.

### DIFF
--- a/dp_creator_ii/__init__.py
+++ b/dp_creator_ii/__init__.py
@@ -1,7 +1,7 @@
 """DP Creator II makes it easier to get started with Differential Privacy."""
 
 import shiny
-from dp_creator_ii.utils.argparse_helpers import get_csv_contrib_from_cli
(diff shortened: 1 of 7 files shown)


ekraffmiller (approved):
I don't know what the problem is that you referred to about using mocks, but this seems to work well to test the app being run with and without the demo data. Approving

PR #121, opened 2024-11-01
What we learned

### LINKED ISSUE #118: Document what we've learned
Even if it seems obvious in retrospect, what questions came up in the application of Python Shiny? This might feed into the getting-started doc for IQSS devs, or maybe back to Shiny itself: It's a new project, and maybe they want feedback?

mccalluc:
- Fix #118 

Will add to this going forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,9 @@ options:
 
 ## Development
 
+This is the first project we've developed with Python Shiny,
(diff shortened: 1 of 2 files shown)


cmbz:
Yes, thanks. Good idea!

cmbz (approved):
Great, thank you. I think that this will be very helpful.

PR #123, opened 2024-11-01
Add a test plan

### LINKED ISSUE #111: Write a test plan
Omer in IQSS may be able to help. Should it be checked into this repo?

mccalluc:
- Fix #111?

These are good questions, but I'm not sure how applicable they are for the project at this phase: Maybe you are requesting a plan for steps before a release? Or there might be a set of IQSS routines that you have in mind, and if this is an IQSS project it might make sense to apply those routines here? Or maybe we want to be more clear about what is expected of reviews? Or maybe we should be more concrete about the feedback we'd like to get from real users? This template seems to assume a separate QA role- I don't think it's my place to dictate staffing for a project, so it's hard for me to fill that out.

So if there's something else I can provide that would be helpful, let me know.

### DIFF
--- a/TEST-PLAN.md
+++ b/TEST-PLAN.md
@@ -5,21 +5,21 @@ Use the Test Plan document to describe the testing approach and overall framewor
 ***Template Instructions***
 *Note that the information in italics is guidelines for documenting testing efforts and activities.  To adopt this template, delete all italicized instructions and modify as appropriate*
 
-# **Introduction** {#introduction}
+# Introduction
 
-##   **Purpose** {#purpose}
+## Purpose
 
 *Provide a summary of the test strategy, test approach, execution strategy and test management.*
 
 Automate the end-to-end functionality of the DP-Creator-II application to ensure it meets user requirements and maintains reliability, especially in the context of privacy-preserving data processing.
 
-##   **Project Overview** {#project-overview}
+## Project Overview
 
 *A summary of the project, product, solution being tested.*
 
-#   **Scope** {#scope}
+# Scope
 
-##   **In-Scope** {#in-scope}
+## In-Scope
 
 *Describes what is being tested, such as all the functions/features of a specific project/product/solution.*
 
@@ -28,7 +28,7 @@ Automate the end-to-end functionality of the DP-Creator-II application to ensure
 \- UI elements related to data privacy settings
  \- Data export functionalities
 
-##   **Out-of-Scope** {#out-of-scope}
+## Out-of-Scope
 
 *Identify all features and combinations of features which will not be tested and the reasons.*
 
@@ -36,9 +36,9 @@ Automate the end-to-end functionality of the DP-Creator-II application to ensure
      \- Non-functional UI aspects not directly related to functionality
      \- External API integrations that are mocked in this environment
 
-#   **Testing Strategy** {#testing-strategy}
+# Testing Strategy
 
-##   **Test Objectives** {#test-objectives}
+## Test Objectives
 
 *Describe the objectives.  Define tasks and responsibilities.*
 
@@ -47,11 +47,11 @@ Automate the end-to-end functionality of the DP-Creator-II application to ensure
    \- Verify UI responsiveness and behavior on supported devices.
    \- Conduct regression testing to maintain code stability with new updates.
 
-##   **Test Assumptions** {#test-assumptions}
+## Test Assumptions
 
 *List the key assumptions of the project and the test plan.*
 
-##   **Data Approach** {#data-approach}
+## Data Approach
 
  *Describe the approach on the test data maintained in QA environments for functional and user acceptance testing.*
 
@@ -59,21 +59,21 @@ Automate the end-to-end functionality of the DP-Creator-II application to ensure
      \- Implement data mocks for complex scenarios.
      \- Develop data validation utilities to check outputs.
 
-##   **Automation Strategy** {#automation-strategy}
+## Automation Strategy
 
      \- Smoke Testing: Quick, high-level test cases to ensure critical paths work.
      \- Regression Testing: Re-running existing tests to validate the latest changes.
      \- E2E Testing: Testing all primary workflows from start to finish.
 
-##   **Test Case Prioritization** {#test-case-prioritization}
+## Test Case Prioritization
 
      \- High priority for critical features like data processing workflows and privacy configurations.
      \- Moderate priority for UI/UX elements like buttons, forms, and navigation.
      \- Lower priority for non-critical paths or backend validations covered in unit tests.
 
-# **Execution Strategy** {#execution-strategy}
+# Execution Strategy
 
-##   **Entry Criteria** {#entry-criteria}
+## Entry Criteria
 
 * *The entry criteria refer to the desirable conditions in order to start test execution*
 * *Entry criteria are flexible benchmarks. If they are not met, the test team will assess the risk, identify mitigation actions and provide a recommendation.*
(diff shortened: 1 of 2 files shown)


TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Introduction
+
+### Purpose
+
+Automate the end-to-end testing of the application to ensure it meets user requirements and maintains reliability, especially in the context of privacy-preserving data processing.
+
+### Project Overview
+
+The single user application will be installed with pip, and offer a web application (running on localhost) that guides the user through the application of differential privacy on a data file they provide.
```
pdurbin, line 11: "Single user application" is a phrase I haven't heard much. I think I know what you mean but perhaps this could be simplified to just "application".

TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+
+- Starting the application with and without `--demo` and other CLI flags.
+- Form filling and navigation between tabs.
+- Conditionally disabled controls (for example, controls late in the flow when inputs have not been provided, and controls early in the flow after the release has been made).
+- Report export.
+- Runnability of any generated code.
+
+### Out-of-Scope
+
+- Performance testing: Test timeouts may incidentally identify components which are slow ([issue](https://github.com/opendp/dp-creator-ii/issues/116)), but it's not the focus of automated testing.
```
pdurbin, line 25:
```suggestion
- Performance testing: Test timeouts may incidentally identify components which are slow ([issue](https://github.com/opendp/dp-creator-ii/issues/116)), but performance is not the focus of automated testing.
```

With "it's" I get slightly confused. I don't think you're referring to "test timeouts", which would be "they're".

TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+- Rendering of preview charts: Might [add results table](https://github.com/opendp/dp-creator-ii/issues/122) to complement graphs, but we're not going to try to make any assertions against images.
```
pdurbin, line 26:
```suggestion
- Rendering of preview charts: We might a [add results table](https://github.com/opendp/dp-creator-ii/issues/122) to complement graphs, but we're not going to try to make any assertions against images.
```

Or plural? Results tables? 🤔

TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+- Testing exact values: The outputs is randomized, so we can not test for any particular value in the output.
```
pdurbin, line 27:
```suggestion
- Testing exact values: The outputs are randomized, so we can not test for any particular value in the output.
```

TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+- Any CSVs used for testing will be checked in to the fixtures directory.
+- Outputs will be checked for structure, but we will not expect any particular values.
+
+### Automation Strategy
+
+- Doc tests: For simple functions, use doctests so the developer can just work with one file.
+- For more complex functions and operations, use pytest unit tests.
+- Run linting and type checking tools as part of tests.
+- Use Playwright for end-to-end tests.
+- Use test matrix on Github CI if we think there will be any senstivity to versions or browsers.
```
pdurbin, line 56:
```suggestion
- Use test matrix on Github CI if we think there will be any sensitivity to versions or browsers.
```

TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+
+N/A: Tests are run automatically on every PR.
+
+### Exit criteria
+
+N/A: Tests are run automatically on every PR.
+
+### Validation and Defect Management
+
+N/A: PRs should not be merged with failing tests.
```
pdurbin, line 74: I'm confused what all these N/A's are here. The sentences themselves sound fine and useful.

Also, I suppose you should make "criteria" Title Case to match the other one.

TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Issue Submission and Prioritization
+
+[See README](https://github.com/opendp/dp-creator-ii/#conventions).
+
+### Pull Request Review and Testing
+
+- Reviewers should read and understand code.
+- Reviewers are not expected to checkout or manually test code.
+
```
pdurbin, line 86: I suggested this [in Slack](https://iqss.slack.com/archives/C07V29323FA/p1730996033997839) but perhaps we could have a bullet for the following idea:

Should we have an additional column after "in review"? For Dataverse after someone clicks "approve" automation moves the PR to the next column. We call this column "ready for QA".

pdurbin, line 86: Can we have a bullet to say "- Reviewers should indicate they are reviewing a PR by dragging it to "In Review" and assigning it to themselves."?

TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+N/A: PRs should not be merged with failing tests.
+
+## Github Workflow
+
+### Issue Submission and Prioritization
+
+[See README](https://github.com/opendp/dp-creator-ii/#conventions).
+
+### Pull Request Review and Testing
+
```
pdurbin, line 83: These comments are all out of order, sorry, but can we have a bullet for "- Reviewers will indicate that that they need the creator of the PR to address a concern by assigning the creator to the issue".

(This would happen while the issue is in the "In Review" column. That's what we do for Dataverse, anyway, and it seems to work fine. The creator unassigned themselves after they've changed the code or at least written a reply to the concern.)

TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+- Reviewers should read and understand code.
+- Reviewers are not expected to checkout or manually test code.
+
+### Continuous Integration and Deployment
+
+Github CI will run tests on each PR. We require tests to pass before merge.
+
+We do not require branches to be up to date with main: There is a small chance that one PR may break because of changes in a separate before, but that can be addressed after the fact.
```
pdurbin, line 91:
```suggestion
We do not require branches to be up to date with main: There is a small chance that one PR may break (have merge conflicts) because of changes in a separate before, but that can be addressed after the fact.
```

By "break" I assume you're talking about merge conflicts.

TEST-PLAN.md
```diff
@@ -0,0 +1,109 @@
...
+
+N/A
+
+## Significantly Impacted Division/College/Department
+
+N/A
+
+## Dependencies
+
+N/A
```
pdurbin, line 109: I'm confused by N/A here. Should it be TBD instead? 🤔

pdurbin (commented):
Overall, seems like a solid plan. I left some comments.

mccalluc:
@anniewu332 thought that the developer of a project shouldn't also be leading the testing: I have my blind spots. I'm going to move this back to draft, and when the right person is identified, they can pick this up.

TEST-PLAN.md
ekraffmiller: I agree with @pdurbin that it would be good to follow the process we use in Dataverse, using an 'In Review' status, and assigning the reviewer/coder if there are changes to be made during the review.

TEST-PLAN.md
ekraffmiller: Personally, I find checking out and running the code helpful to understanding it, so I usually do it.  I don't think we need to say that reviewers are not expected to do that.  Also, since we don't have a QA step, running the code helps to find bugs in the review step.

TEST-PLAN.md
ekraffmiller: I think it's a good practice to have PRs up to date with main before they are reviewed, especially if there are merge conflicts.  Or, if a PR depends on another branch, it can point to that branch instead of main.

TEST-PLAN.md
ekraffmiller, line 27: Is it possible to test that the outputs are within a certain range, just as a sanity check?

ekraffmiller (commented):
This looks good, thank you, I just left some comments about the details

PR #126, opened 2024-11-02
Disable buttons until required inputs supplied

### LINKED ISSUE #124: "Download results" button disabled until at least one column is selected (and explain why disabled)

### LINKED ISSUE #125: "Define analysis" button disabled until file is selected.

mccalluc:
- Fix #125 
- Fix #124

### DIFF
--- a/dp_creator_ii/app/dataset_panel.py
+++ b/dp_creator_ii/app/dataset_panel.py
@@ -29,6 +29,7 @@ def dataset_ui():
             "contributions",
             ["Contributions", ui.output_ui("contributions_demo_tooltip_ui")],
             contributions,
+            min=1,
         ),
         ui.output_ui("python_tooltip_ui"),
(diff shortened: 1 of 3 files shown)


dp_creator_ii/app/dataset_panel.py
ekraffmiller, line 1: I was able to clear the Contributions input, and the "Go to Analysis" button was still enabled.  So I think that needs to be checked too, right?

ekraffmiller (dismissed):
Looks good, one change request

dp_creator_ii/app/dataset_panel.py
mccalluc: Good catch. That prompted me change both of these to reactive calcs, rather than trying to update the state in two separate event handlers.

ekraffmiller (approved):
I tried it and the validation is working as expected 👍

PR #134, opened 2024-11-04
lower/upper instead of min/max

### LINKED ISSUE #133: "Lower"/"Upper" instead of "Min"/"Max"
Fix the names: It's a small distinction, but we're not asking the user to guess the min and the max -- what we need are bounds, and the actual min or max may certainly be outside.

mccalluc:
- fix #133


mccalluc:
Put this on hold: Get the reactives at the top level into man, and then see what is still needed.

mccalluc:
- integrated in #142. Closing.

PR #135, opened 2024-11-04
Demo tooltips on the analysis panel

### LINKED ISSUE #127: Demo tooltips on Define Analysis
on checkboxes:
> Not all columns need analysis. For this demo, just check "class_year" and "grade". With more columns selected, each column has a smaller share of the privacy budget.

on epsilon slider:
> If you set epsilon above one, you'll see that the distribution becomes less noisy, and the confidence intervals become smaller... but increased accuracy risks revealing personal information.

on min/max:
> DP requires that we limit the sensitivity to the contributions of any individual. To do this, we need an estimate of the lower and upper bounds for each variable. _We should not look at the data when estimating the bounds!_ In this case, we could imagine that "class year" would vary between 1 and 4, and we could limit "grade" to values between 50 and 100.

on bins:
> Different statistics can be measured with DP. This tool provides a histogram. If you increase the number of bins, you'll see that each individual bin becomes noisier to provider the same overall privacy guarantee. For this example, give "class_year" 4 bins and "grade" 5 bins.

on weight:
> You have a finite privacy budget, but you can choose how to allocate it. For simplicity, we limit the options here, but when using the library you can fine tune this.

mccalluc:
- Fix #127

I think something like this would be good... but maybe this is too much text for this format?


ekraffmiller (dismissed):
looks good, I like all the information about how to use the demo data.  Maybe in a future iteration, it could be made more prominent, maybe as extra text on the page, rather than in a tooltip.  But I thinks it works well for now.

Issue #136, opened 2024-11-05
Disable weight if only one column selected


mccalluc:
If there's only one column, changing the weight doesn't do anything, which is confusing. Should disable this with an explanation when there is only one column selected.

PR #137, opened 2024-11-05
Two column layout: Graph expands to fit

### LINKED ISSUE #131: two-column layout, with inputs on the left, and the chart on the right
- See also #132

### LINKED ISSUE #132: key-value pairs on one line on analysis tab
Each key-value should be together on a line.

- See also #131

mccalluc:
- Fix #131 
- Close #132 without fix: There is plenty of vertical space, and it's not immediately obvious from the docs how to get the label and the input on the same line.
- Demoing for Annie, and she would prefer a wider page margin, so I tweaked that. I still hope there is a "right" way to do this that extends the tab line to the edge of the window, but I don't have that solution right now.

I would also like to avoid repeating the warning above the graph, but that's more of a logic change, so don't worry about it right now.
- #138

Working with this part of the UI prompted me to file a new issue:
- #136

Time required: ~1 hour


ekraffmiller (approved):
looks good!

PR #142, opened 2024-11-05
Move reactive values up to top level

### LINKED ISSUE #129: Reactive values for analysis details should live at the top level
Analysis details need to live at the top level, so they can be used to fill in the blanks for the analysis.

### LINKED ISSUE #133: "Lower"/"Upper" instead of "Min"/"Max"

mccalluc:
- Fix #129
- Incorporates #134 / fixes #133

Time: ~2.5 hours


ekraffmiller (commented):
I tested this, and when columns are added in the Analysis Tab, the inputs are reset, see example:

https://github.com/user-attachments/assets/6ebd8366-0a1c-4be4-a7f1-ad7e57fa1de7

ekraffmiller (changes requested):
Found an issue when checking a column on the Analysis page, see comments

ekraffmiller (approved):
Looks good, values are retained when new columns are checked.

Issue #144, opened 2024-11-05
Options for visualization uncertainty, other than whiskers


mccalluc:
From Mark Keller at HMS:

> That is an interesting problem, I am not sure about visualization for differential privacy specifically but it definitely is related to uncertainty visualization. Your sketches remind me of the ggdist R package. There is a publication by the same author titled When(ish) is my bus: 
> https://www.mjskay.com/papers/chi_2016_uncertain_bus.pdf and this recent paper is on Uncertainty for Set visualization but contains a nice Table 1 that lists papers related to uncertainty visualization in general: 
> https://drive.google.com/file/d/1xewFIlurPA3_bJoBXxtRenc6MNmIHn3O/view?usp=drivesdk
> I am not familiar with all of them but it could be a good starting point.

PR #145, opened 2024-11-05
Set bottom of axis

### LINKED ISSUE #143: Plot should not go below zero
We know that negative counts are one thing that weirds people out when they are new to DP. One thing we can do to mitigate this is set the lower bound of the chart to zero. Any whiskers will still peak above.

mccalluc:
- Fix #143

Time: ~0.25 hours


ekraffmiller (commented):
This works as described, but I wonder if it might be good to keep the negative whiskers as a way of teaching how DP works.  I think it's an important aspect of it.  But that's just my $.02!

mccalluc:
I don't think it's useful to keep them. I do think we need to explain the whiskers, and I've filed 
- #144

to explore other ways of presenting uncertainty, but I don't think the negative axis does this.

ekraffmiller (approved):

PR #146, opened 2024-11-05
Cross-platform safe temp file

### LINKED ISSUE #141: `/tmp` might not be portable
I'm not sure how well this will run on windows, but a first step would be to use a proper temp file instead of `/tmp`.

mccalluc:
- Fix #141 
- On the good side, this is more likely to work on Windows
- On the bad side, a new temp directory is created on every run
  - We could clean it up in the `session.on_ended` handler, but I'm not sure the complexity is worth it?

time: ~1 hour, mostly thinking about different options.

### DIFF
--- a/dp_creator_ii/utils/argparse_helpers.py
+++ b/dp_creator_ii/utils/argparse_helpers.py
@@ -5,6 +5,7 @@ import csv
 import random
 from warnings import warn
 from collections import namedtuple
+from tempfile import mkdtemp
 
 
 def _existing_csv_type(arg):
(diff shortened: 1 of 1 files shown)


ekraffmiller:
As another option, could we use [gettempdir()](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir)?  It supports different environments.  
Or, can we create demo.csv just once and include it in the application?

mccalluc:
Good call!

ekraffmiller (changes requested):
See comment about using a temp dir

mccalluc:
It looks like that is still doesn't return something as simple as `/tmp`:
```
>>> from tempfile import gettempdir
>>> gettempdir()
'/var/folders/jm/40wztt291610f8pk5w0sw20r0000gq/T'
```
I'd prefer to keep the status quo: I think it's nice to have a directory that is just for this application, rather than sharing one that could inadvertently interfere with other applications.

mccalluc:
Second, or third thought: Would be good to try checking in the generated file. One the fence, but it would mitigate this whole temp directory problem.

mccalluc:
- Close in favor of #167

PR #147, opened 2024-11-06
Handle empty column headers

### LINKED ISSUE #80: Empty column header in CSV causes error
We shouldn't do anything special, but we should check that empty strings work in the UI.

mccalluc:
- Fix #80
- also handle duplicate columns,
- and columns that are only distinguished by non-word characters

Time: ~4 hours


ekraffmiller (approved):
tested with different column headings, looks good

PR #148, opened 2024-11-06
Do not label every bin

### LINKED ISSUE #128: Above 8 bins, don't label every bin
With lots of bins, the labels overlap. Maybe at 8 bins just label every other bin. (And at 16, every third? But you probably shouldn't have that many bins.)

mccalluc:
The exact cutoff might change.

- Fix #128

Time: ~1 hour


ekraffmiller (approved):
Tested, and I think it looks good in most cases.

PR #152, opened 2024-11-08
Make notebook with plots for columns

### LINKED ISSUE #36: Think about notebook execution
Having outputs in the Jupyter notebook is an interesting idea, but if we ever do it we'd want to make sure it was the same DP release as other possible downloads, so it didn't burn another epsilon. For simplicity, for now, drop it, and remove supporting dependencies (ipykernel, at least).

### LINKED ISSUE #42: Explain `utf8-lossy` in templated code
In this PR
- https://github.com/opendp/dp-creator-ii/pull/29

we decided to use `utf8-lossy` to load data, but we should explain that choice to downstream users in the templated code.

### LINKED ISSUE #46: Notebook to demo DP histograms with cut
Before trying to integrate it into the codebase, try to use [`cut`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.cut.html) to generate DP histograms, and in a notebook for now. Down the road, we can pull this out.

mccalluc:
- Fix #46 
- Fix #42 
- Fix #36 

Dropped the expected-script test: when it was short, it was useful, but now it just means that every change requires a corresponding update to the test: Better to have finer grained unit tests that catch problems earlier.

In the notebook template, added lots of comment blocks that will be turned into markdown cells. The `# +` and `# -` are the [jupytext light format](https://jupytext.readthedocs.io/en/latest/formats-scripts.html#the-light-format).

This is a change, but I think that having the executed notebook be the real release from the perspective of this application makes sense: That's the best way of making sure the code in the notebook does what we think it should do.

For reviewer:
- This would be a good one to checkout and run end to end. (With the name change, you'll need to run `flit install` before you can execute `dp-wizard`.)
- There is a lot undone, so don't worry about gaps, but if something is just wrong, that would be good to hear about.
- After this, I want to take some time to clean up the internals, but there won't be big changes in the UI... so if something would be really good to get sorted before the demo next week, that would be good to note.


ekraffmiller:
Looks good, I started from scratch with the new repository and downloaded a notebook with histograms created.
One thing I noticed, if I change the number of bins using the arrows, then the bins appear out of order in the histogram plot.  This happens in the app and in the notebook.
For example,

Screenshot 2024-11-15 at 11 32 05 AM

Also, if I click a column and then unclick it, the column appears in the results page. For example, in this test, I clicked and unclicked the sex column:

Screenshot 2024-11-15 at 12 06 21 PM

ekraffmiller (commented):
Looks good, I just noticed that histogram bins are sometimes out of order, see below.  For this PR, should I test the generated python script, or is that still a work in progress?

mccalluc:
> bins appear out of order in the histogram plot

Good catch: These strings are being sorted alphabetically, so "10" comes before "2". I think the best way to fix this is to pull out the first number of a pair and sort, probably on the python side, rather than trying to parse the string inside polars. Will file a new issue.

mccalluc:
> should I test the generated python script, or is that still a work in progress?

Not really necessary right now. There's going to be more work on these.

mccalluc:
Going ahead and fixing the sorting...

mccalluc:
Now also
- Fix #165 
- Fix #163

ekraffmiller (approved):
looks good! bins sorting correctly

Issue #158, opened 2024-11-13
Rename: dp-creator-ii -> dp-wizard!


mccalluc:
- find and replace in project
- update the repo
- emoji? 🧙‍♂️

mccalluc:
repo renamed! this is done.

PR #161, opened 2024-11-14
creator ii -> wizard

mccalluc:
- Towards #158 
- Keep open until repo rename is complete

The only remaining reference to "creator" is correct:
```
$ grep -ri --exclude-dir .venv --exclude-dir __pycache__ --exclude-dir .mypy_cache --exclude-dir .git --exclude-dir dist --exclude-dir .pytest_cache creator .
./README.md:Building on what we've learned from [DP Creator](https://github.com/opendp/dpcreator), DP Wizard will offer:
```

For reviewer:
- I've done a google check, but confirm that "DP Wizard" feels like a good name.


pdurbin (approved):
I ran `ack -i creator` and only found found this, which is fine:

`Building on what we've learned from [DP Creator](https://github.com/opendp/dpcreator), DP Wizard will offer...`

Approved! 🧙 🪄

PR #167, opened 2024-11-19
Write to a relative temp dir

### LINKED ISSUE #141: `/tmp` might not be portable

mccalluc:
- Fix #141 
- Replace #146

Might also use this for uploaded files. Currently if you upload a file through the UI, the path is a really obscure temp directory, and that's what we end up showing in the notebook. A scratch space that's local to the code might be a better solution.

### DIFF
--- a/dp_wizard/tmp/.gitignore
+++ b/dp_wizard/tmp/.gitignore
@@ -1,0 +1,1 @@
+demo.csv
--- a/dp_wizard/utils/argparse_helpers.py
+++ b/dp_wizard/utils/argparse_helpers.py
@@ -81,10 +81,10 @@ def _get_demo_csv_contrib():
     """
     random.seed(0)  # So the mock data will be stable across runs.
 
-    csv_path = "/tmp/demo.csv"
+    csv_path = Path(__file__).parent.parent / "tmp" / "demo.csv"
     contributions = 10
 
-    with open(csv_path, "w", newline="") as demo_handle:
+    with csv_path.open("w", newline="") as demo_handle:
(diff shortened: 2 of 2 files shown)


pdurbin:
I see what you mean. On "main" we get this:

`"    data=pl.scan_csv('/var/folders/wk/180l5f012g1_zjbrhsxgw8dw0000gp/T/fileupload-izsvgh4k/tmpse6ja5ip/0.csv', encoding=\"utf8-lossy\").with_columns(grade_config),\n",`

pdurbin:
Huh, maybe I'm not testing this right but on "141-temp-in-repo" I'm getting the same thing:

`"    data=pl.scan_csv('/var/folders/wk/180l5f012g1_zjbrhsxgw8dw0000gp/T/fileupload-4j1j74yc/tmptytthuqq/0.csv', encoding=\"utf8-lossy\").with_columns(grade_config),\n",`

@mccalluc am I looking in the right place?

mccalluc:
@pdurbin : Sorry I wasn't clear about the scope. "_Might_ also use this for uploaded files." This PR right now is just trying to fix the non-portable use of `/tmp`, but if this much seems good, I could extend it, preferably in a follow-up PR.

pdurbin (approved):
When I run with ``--demo`` everything still works. I see "demo.csv" and can walk through the app. Approved.

Screenshot 2024-11-19 at 2 54 40 PM

PR #168, opened 2024-11-19
OO-ify the code generation code

### LINKED ISSUE #156: In template, error if code block substitution fails
If the placeholder for a block substitution is not alone on a line, the substitution will fail, and the only error will be when the string at the end has an un-substituted placeholder.

### LINKED ISSUE #162: More OO for templates
We have:
- `make_notebook_py`
- `make_script_py`

and
- `_make_context_for_notebook`
- `_make_context_for_script`

with a lot duplicated between the pairs of functions. Try to refactor this in a more OO way, with the shared logic in a super class.

mccalluc:
- fix #162 and clean up a lot of duplicated code
- fix #156: Stonger tests of template filling
- Rename from `template` to `code_generators`: the templates are just a means to an end.
- ... and put the template machinery in its own file.
- Named tuples instead of dicts for `AnalysisPlan` and `AnalysisPlanColumn`: Complements #164

For the reviewer:
- The diff is going to be hard to read, but if you just look at the result, does the code organization make sense?
- Are there other obvious failure modes that should be tested on the templates?


pdurbin (approved):
I didn't run the code but the new version looks fine to me.

PR #169, opened 2024-11-20
Explain confidence interval

### LINKED ISSUE #153: Explain confidence interval
At least explain that it is 95%; May want to let user determine the confidence interval, or display multiple confidence intervals?

mccalluc:
- Fix #153.
- My sense is that the confidence interval is not something we want to control in the UI: It's the kind of detail that is best left for the user to tweak in the notebook.

For the reviewer:
- Does a short comment in the UI (where we want to avoid distractions) and a longer note in the notebook (where we want to really explain things) make sense?
- Do you think explanation is needed anywhere else? Could add a comment to the graphs, but I'd worry about clutter.


ekraffmiller (approved):
I think short explanation in the generated code looks good. We can wait for user feedback to see if more is needed.

PR #170, opened 2024-11-20
Hide weight if only one column selected

### LINKED ISSUE #136: Disable weight if only one column selected

mccalluc:
- Fix #136

Shiny doesn't seem to support disabling select controls. Not in the docs, and just adding a kwarg doesn't work. Adding a note after the control follows the pattern we have with disabled buttons. For both of these styling it as a [bootstrap alert](https://getbootstrap.com/docs/4.0/components/alerts/) might be appropriate.

For reviewer:
- Does this message seem worthwhile? or maybe it's a distraction if we can't easily disable the select?
- Would styling this message as an alert be appropriate? or would that call too much attention to it?

On the whole I think this is a good change, but I'm not totally confident about it. Feedback welcome.

---

Update: Instead of adding a message, just use a hidden div when there is only one column selected. PR retitled.

### DIFF
--- a/WHAT-WE-LEARNED.md
+++ b/WHAT-WE-LEARNED.md
@@ -2,6 +2,10 @@
 
 Even if it seems obvious in retrospect, what have we learned about Python Shiny in this project?
 
+## Some basic html attributes are not supported
+
+I can mark a button as disabled, but it doesn't seem that a `ui.input_select` can be disabled.
+
 ## No warning if ID mismatch / type mismatch
 
 Unless I'm missing something, there doesn't seem to be any warning when there isn't a matching function name in the server for an ID in the UI. Either from typos, or fumbling some more complicated display logic, there have been times where this could have been helpful.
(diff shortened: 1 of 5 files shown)


ekraffmiller (commented):
I tried it out, and it's a little confusing to have the input still enabled, especially since changing the value changes the histogram, although I think that's just randomness of the generated bins. 
Is it possible to remove the input select altogether, when there is only one variable chosen?

mccalluc:
@ekraffmiller : Good thoughts. I tend to shy away from controls that are disabled based on some other criterion. We've all been in the place where we can see the option we need, and it's so close, but it's grey for no apparent reason.

I'm going to try a new PR, but instead of adding a message, just wrapping the control in a hidden div. Stay tuned...

ekraffmiller (approved):
Looks good!

PR #171, opened 2024-11-20
Info boxes about locking panels (without actually doing it)

### LINKED ISSUE #88: Lock page to changes
## Overview
- This functionality enables locking of page changes to prevent users from altering parameters such as their privacy budget

## Dependencies
- None

## Related
- https://github.com/opendp/dp-creator-ii/issues/86

mccalluc:
- Towards #88
- Keep track of which panel we _should_ be on, even if we clicked on a tab to look at another panel.
- Explain why the user won't be allowed to touch these other pages
- Wrap the message in a bootstrap info box.

There's enough here that I'd like sign off from @Shoeboxam and someone else before going farther.

For reviewers:
- For both Mike and someone else: Do the warning messages express well why folks shouldn't be on the other tabs?
- Is the styling ok? I'm not intending to spend a lot of time tweaking the UI, but do you think this is adequate? Maybe the boxes are too much, and bold text would suffice?

### DIFF
--- a/dp_wizard/app/__init__.py
+++ b/dp_wizard/app/__init__.py
@@ -11,7 +11,6 @@ logging.basicConfig(level=logging.INFO)
 
 app_ui = ui.page_bootstrap(
     ui.head_content(ui.include_css(Path(__file__).parent / "css" / "styles.css")),
-    ui.output_text("current_panel"),
     ui.navset_tab(
         dataset_panel.dataset_ui(),
         analysis_panel.analysis_ui(),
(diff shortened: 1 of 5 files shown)


ekraffmiller (commented):
I think the explanatory text is good, it describes well why the user shouldn't go back and tweak the inputs.  
One comment on the locking behavior - could we make the lock only appear if the user has actually downloaded the release, as opposed to just viewing the download page?

I'm leaving this as a comment rather than approving, because @mccalluc mentioned that he would like @Shoeboxam's feedback on this.

mccalluc:
> could we make the lock only appear if the user has actually downloaded the release, as opposed to just viewing the download page?

Right now the final tab is just buttons, but my sense is that we may want to include a summary of the results on that page itself... perhaps the graphs? so the release would happen when you enter that tab by clicking the submit button on the previous tab. I think consistency in the locking behavior would be good: When you submit your selections on one page they are fixed, whether that is the first or the second tab.

mccalluc:
@ekraffmiller, I'll wait for Mike's thoughts, so probably better to just to leave this as a comment so I don't merge it by mistake. Thanks for the feedback.

mccalluc:
I'll keep the issue open, but my sense is that this isn't so important right now. Even UI changes like right aligning the buttons will emphasize the flow, and as a single user application, we can't really stop people from abusing their privacy budget, in any case.

PR #181, opened 2024-11-22
Text (yaml) report for download

### LINKED ISSUE #110: Return results from execution of templated code
We want to make sure we're eating the dogfood, so we'd like the preview calculations
- (ie, in #87)

to use the same code that will eventually be included in the template. Probably `exec_` functions to complement the `make_` functions, where needed? and just return a dict of the namespace?

### LINKED ISSUE #173: Download Text report

### LINKED ISSUE #177: "Please wait" while notebook builds?
Even with one column it's long enough that it feels like nothing is happening. Maybe there's a way to send the first byte faster so the browser knows a download is starting? Or build and run the notebook on entering the download page, instead of waiting for button press?

mccalluc:
- Fix #173
- Fix #177
- Fix #110

For reviewer:
- Is the template code understandable?
- Is the generated code in the notebook ok?
- Is the format of the report ok? (Not expecting you to mind-read hypothetical future users, but anything that should just be fixed now?)

### DIFF
--- a/WHAT-WE-LEARNED.md
+++ b/WHAT-WE-LEARNED.md
@@ -2,6 +2,10 @@
 
 Even if it seems obvious in retrospect, what have we learned about Python Shiny in this project?
 
+## Some basic html attributes are not supported
+
+I can mark a button as disabled, but it doesn't seem that a `ui.input_select` can be disabled.
+
 ## No warning if ID mismatch / type mismatch
 
 Unless I'm missing something, there doesn't seem to be any warning when there isn't a matching function name in the server for an ID in the UI. Either from typos, or fumbling some more complicated display logic, there have been times where this could have been helpful.
--- a/dp_wizard/app/analysis_panel.py
+++ b/dp_wizard/app/analysis_panel.py
@@ -113,6 +113,7 @@ def analysis_server(
                 bin_counts=bin_counts,
                 weights=weights,
                 is_demo=is_demo,
+                is_single_column=len(column_ids) == 1,
             )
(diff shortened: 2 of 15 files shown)


ekraffmiller (commented):
For the report, it looks good, but I think it would be helpful for the user if all the imports for the release were listed. 
For example, this is the generated report I got:
```
inputs:
  data: /var/folders/sv/ps25r8d562z0jzv4y4c_yn7c0000gq/T/fileupload-f2gfx9jw/tmppn4cvg9x/0.csv
outputs:
  educ:
    accuracy: 3.375617766595713
    confidence: 0.95
    histogram:
      (-inf, 1]: 33
      (1, 2.5]: 14
      (10, 11.5]: 167
      (11.5, 13]: 254
      (13, 14.5]: 53
      (14.5, 16]: 36
      (2.5, 4]: 57
      (4, 5.5]: 24
      (5.5, 7]: 54
      (7, 8.5]: 47
      (8.5, 10]: 261

```
I think it would be good to also include, the epsilon, weight, and maybe number of bins? (Although that is shown in the actual result)

mccalluc:
I think this adds all the required info, but we're just inserting the values into the generated code: The values also occur earlier in the template, but they aren't stored into variables. Doing that would add a level of generality that would interfere with the clarity of the examples at the top of the notebook, which I think are the priority... but I could be wrong.

ekraffmiller (approved):
I think it looks good, I agree that you don't need to change the notebook to put the values into variables.

PR #182, opened 2024-11-22
CSV report

### LINKED ISSUE #172: Download CSV report

mccalluc:
- Fix #172
- also put all the download buttons in a grid with icons

Not committed to this format for the reports: Just heard that there should be a CSV, but there don't seem to be be any requirements beyond that.


ekraffmiller (approved):
I think this looks reasonable, I just have the same comment I have for the report that I think it would be good for the user to see the other inputs (epsilon and weight)

mccalluc:
Thanks: Since this is approved, I'll merge it to `173-text-yaml-report`, address the comments there, and request re-review.

PR #183, opened 2024-11-24
Let user set number of rows in simulated data

### LINKED ISSUE #154: Does the number of rows in the fake data impact visualization?
Should we prompt the user for an order of magnitude estimate for their data, so the previews have plausible confidence intervals?

mccalluc:
- Fix #154

For the reviewer:
- Does the introduction to this option make sense?
- Is this a good set of choices? I was hoping that thinking in terms of orders of magnitude lessens the pressure on the user when coming up with an estimate, and that 10 is too small to be useful, while 100,000 is so large that the improvement is negligible, leaving us with 100, 1,000, and 10,000.

### DIFF
--- a/dp_wizard/app/analysis_panel.py
+++ b/dp_wizard/app/analysis_panel.py
@@ -28,6 +28,16 @@ def analysis_ui():
         ),
         ui.output_ui("columns_ui"),
         ui.markdown(
+            "What is the approximate number of rows in the dataset? "
+            "This number is only used for the preview and not the final calculation."
+        ),
+        ui.input_select(
+            "row_count",
+            "Approximate Rows",
+            choices=["100", "1000", "10000"],
+            selected="100",
+        ),
+        ui.markdown(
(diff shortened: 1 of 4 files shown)


ChengShi-1 (commented):
I have no problem with 100, 1,000, and 10,000. Just an idea. It may be cleaner for users if we could make the selection a range instead of a number? Selections may look like 10-100, 100-1000, 1000-10,000 or something like these.

mccalluc:
I don't think that would be more clear.
- It is selecting one particular number: With 100, the simulated dataset has 100 rows, ditto for 1000 and 10000. It's not clear to me whether "100-1000" would be better represented by 100, or 1000, or something in between.
- I want to take the pressure off the user for knowing how many rows are in their dataset. Ideally they don't even look at their data. I worry that putting a range on it will make them think that they should know which range their data is in.

ChengShi-1 (approved):
Visually reviewed the code, and run it as well. Approve cause everything works well

Issue #94, opened 2024-10-24
Improve page UI/UX


cmbz:
## Overview
- Tasks to improve the UI/UX of the `Define Analysis` page

## Issues
- [ ] https://github.com/opendp/dp-creator-ii/issues/82
- [x] #128
- [x] #133
- [x] #136

## Dependencies
- None

## Related
- https://github.com/opendp/dp-creator-ii/issues/93

mccalluc:
Done

PR #184, opened 2024-11-25
Add cards on analysis page

mccalluc:
- Towards #94
- Wrap each column interface in a card.
- Use the space at the top more efficiently, too.
- The formatting of the slider isn't great, but if the overall layout here seems like an improvement, I'd like to leave that for later.

For reviewer:
- This is subjective, but does this seem to present the steps in defining the analysis a little more clearly?


ekraffmiller (approved):
This looks great!  I think the new layout makes everything happening on the page more understandable 👍

PR #191, opened 2024-11-26
Check that it runs out of the box

### LINKED ISSUE #60: Is this a library or an application?
If it's an application, we should aggressively lock dependencies (including in pyproject.toml)... but if it's a library that should play well with other software, we should keep them loose.

Not sure which is the best approach right now.

## Related
- https://github.com/opendp/dp-creator-ii/issues/90

mccalluc:
- Fix #60: For the time being, it's an application... but we want to be sure not to get boxed in by overly strict version requirements.

This is checking two slightly different things:
- Can we run the end-to-end test with only a couple extra packages besides what a user would get when they install this package? We want to be sure that we aren't relying on dev dependencies in the production code by mistake.
- Can we run the end-to-end test with unpinned dependencies? We want to know if newer versions have introduced breaking changes. (We'll keep the pinned dependencies, too, so if something does go wrong with the unpinned deps, we know who is to blame.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,9 +35,10 @@ jobs:
           pip install pytest pytest-playwright
 
       - name: Install browsers
+        # TODO: Why doesn't work to just install chromium now?
         # Install just one browser instead of the default three.
         # https://playwright.dev/python/docs/browsers#managing-browser-binaries
-        run: playwright install chromium
+        run: playwright install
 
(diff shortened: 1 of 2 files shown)


mccalluc:
I think we have some improvements here, for example
- #356

I also think that without containerizing, it's going to be hard to be 100% sure that a runtime dependency isn't coming from the test dependencies, and I think this might have been the case with `webpdf`.

Closing.

PR #199, opened 2024-12-05
just use plain pip install

### LINKED ISSUE #192: Just use `pip install --editable .`?
- see https://github.com/pypa/flit/issues/670

mccalluc:
- Fix #192


ekraffmiller (approved):
I tested running locally with the pip install command, it works as expected, approving

PR #205, opened 2024-12-19
Include unit of privacy in graphs and output

### LINKED ISSUE #202: Show unit of privacy on Analysis tab
Perhaps in the title of the graph?

mccalluc:
- Fix #202 

Started with just the preview graphs, but then realized it should be included in the final outputs, too, though it does add some complexity.


ekraffmiller (approved):
looks good, approved

PR #206, opened 2024-12-19
Remove unused sort

### LINKED ISSUE #178: Remove `sort` after `collect`
It's just sorting the strings, and the real numeric sort happens in `df_to_columns`

mccalluc:
move it out of the code, but add it to the doctest for stability.

- Fix #178


ekraffmiller (approved):
This looks good. When testing I noticed an issue with the display of the histogram in the notebook, added an issue [here](https://github.com/opendp/dp-wizard/issues/210).

PR #207, opened 2024-12-19
Document dependencies in generated code

### LINKED ISSUE #166: With downloaded code, describe the necessary pip installs
In particular `pip install 'opendp[polars]'`
- [ ] For the script, consider zipping it up with a requirements.txt?
- [ ] For the notebook, consider some commented out `pip install`s

mccalluc:
- Fix #166 

Having a separate requirements.txt with the script feels like overkill.

We also don't have a test that confirms that the given dependencies are sufficient, but that also feels like overkill.


ekraffmiller (approved):
looks good, approving

PR #209, opened 2025-01-03
Strip coda from notebook

### LINKED ISSUE #208: Strip cells that generate artifacts from notebook?
The notebook serves two purposes:
- It's an artifact that the user will download and continue to work with
- It is code that generates our other exports (yaml and csv downloads)

What if we were to strip the last cells (those that generate reports) from the notebook? They are a bit of a distraction, and they write to the host filesystem in ways that a standalone notebook usually wouldn't.

mccalluc:
- Fix #208


ekraffmiller (approved):
Looks good, approved

PR #211, opened 2025-01-07
confidence interval text + histogram table

### LINKED ISSUE #122: Tables of results, to complement preview charts
- For accessibility: For some users, a table may be better than a chart.
- For testing: It's easier to make assertions against text than graphics.

### LINKED ISSUE #203: Give the confidence interval explicitly, and not just in the graph
People want that number, and eye-balling it is hard.

mccalluc:
- Fix #203 
- Fix #122

Previously, the generation of the fake data and rendering were in the same function. This pulls the generation of fake data into a separate reactive calc that gets called from several different functions.

For the reviewer:
- Are the UI and layout reasonable?
- The app is getting more complicated, but none of the complexity really has anything to do with DP. Is this codebase going to be maintainable by people other than Chuck?

### DIFF
--- a/dp_wizard/app/__init__.py
+++ b/dp_wizard/app/__init__.py
@@ -36,6 +36,7 @@ def make_server_from_cli_info(cli_info: CLIInfo):
         upper_bounds = reactive.value({})
         bin_counts = reactive.value({})
         weights = reactive.value({})
+        accuracy_histograms = reactive.value({})
         epsilon = reactive.value(1.0)
 
         dataset_panel.dataset_server(
@@ -57,6 +58,7 @@ def make_server_from_cli_info(cli_info: CLIInfo):
             upper_bounds=upper_bounds,
             bin_counts=bin_counts,
             weights=weights,
+            accuracy_histograms=accuracy_histograms,
             epsilon=epsilon,
         )
         results_panel.results_server(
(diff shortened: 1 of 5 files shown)


mccalluc:
> code for generating the table is not included as a code snippet in the analysis tab

You mean the fake, normally distributed data? That's true, but I would think that would be a distraction: In the notebook, the real data is being read, so there's no need for this fake data. But if we hear something different from users, could consider it.

ekraffmiller:
> > code for generating the table is not included as a code snippet in the analysis tab
> 
> You mean the fake, normally distributed data? That's true, but I would think that would be a distraction: In the notebook, the real data is being read, so there's no need for this fake data. But if we hear something different from users, could consider it.

Ok, I just thought it would help the users see how the histogram that is based on the fake data is being generated, since it's the same process that will be used on the real data. We can see what the users think 👍

mccalluc:
> see how the histogram that is based on the fake data is being generated

Think I understand now. You're right, `dp.Context.compositor` is used in two separate places:
- `dp_helper.py` (used by the Shiny UI)
- `_context.py` (template used to generate notebook)

Two TODOs:
- eventually combine these, so we know they are in sync
- Also: use `_context.py` to make a code snip for the user (although that is the entire computation, and maybe we just want to look at one stat at a time.)

PR #212, opened 2025-01-07
confidence interval in plot module

mccalluc:
It would probably make sense to pull all the histogram preview logic into a separate Shiny module, but it's not working right now, and this refactoring probably doesn't make the base PR easier to review. Making a draft PR to save the work, but not sure it's a priority to finish this.

- Extends #211


mccalluc:
This was handled by 
- #211

Closing.

PR #213, opened 2025-01-10
Maybe changing the aspect ratio helps?

### LINKED ISSUE #210: Column labels for histogram are too close in generated Notebook
In the app,  when the labels for histogram buckets can't fit on the histogram, some labels are skipped, for example: 

Screenshot 2025-01-07 at 3 22 55 PM

This isn't happening in the Notebook, so the labels aren't legible:

Screenshot 2025-01-07 at 3 22 33 PM

Can we add the same logic to the Notebook, to skip some labelling when the labels are too close together?

mccalluc:
- fix #210

It's going to be hard to get a simple solution that works in all cases, but this might help.

### DIFF
--- a/dp_wizard/utils/shared.py
+++ b/dp_wizard/utils/shared.py
@@ -47,6 +47,8 @@ def plot_histogram(
     """
     import matplotlib.pyplot as plt
 
+    plt.rcParams["figure.figsize"] = (12, 4)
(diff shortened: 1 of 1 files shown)


ekraffmiller (approved):
great, that definitely looks better 👍

PR #214, opened 2025-01-10
Validate contributions

### LINKED ISSUE #188: Input validation  / use `req`
Clearing out the numeric fields can cause errors from the framework in the UI. Probably just `req`?

mccalluc:
- Fix #188

The user could also misfill Upper/Lower/Bins on the next tab, but in that case, there is an immediate error message, albeit an ugly one, so I'm inclined not to add more validation code there. 

For reviewer:
- Is this UI good enough?
- Ok not to investigate test flakiness (`define_analysis_button.is_disabled()`) more deeply?
- Confirmation that we're ok with Shiny's default error messages, where the error is obviously connected to the input?

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -68,6 +68,7 @@ def test_default_app(page: Page, default_app: ShinyAppProc):  # pragma: no cover
     # Check validation of contributions:
     # Playwright itself won't let us fill non-numbers in this field.
     # "assert define_analysis_button.is_enabled()" has spurious errors.
+    # https://github.com/opendp/dp-wizard/issues/221
     page.get_by_label("Contributions").fill("0")
     expect_visible("Contributions must be 1 or greater")
     expect_visible("Choose CSV and Contributions before proceeding")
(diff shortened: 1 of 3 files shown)


ekraffmiller (commented):
I think the validation looks good, the message refers to the input field, so it is clear.  I think it's ok to leave the other validations and the flaky test for later, since the UI is still evolving. But it would be good to add an issue or a TODO in the code as a reminder to review it later.

tests/test_app.py
```diff
@@ -64,6 +64,18 @@ def expect_no_error():
     # Now upload:
     csv_path = Path(__file__).parent / "fixtures" / "fake.csv"
     page.get_by_label("Choose CSV file").set_input_files(csv_path.resolve())
+
+    # Check validation of contributions:
+    # Playwright itself won't let us fill non-numbers in this field.
+    # "assert define_analysis_button.is_enabled()" has spurious errors.
```
mccalluc, line 70:
```suggestion
    # "assert define_analysis_button.is_enabled()" has spurious errors.
    # https://github.com/opendp/dp-wizard/issues/221
```

mccalluc:
@ekraffmiller , filed an issue, and added link in the comment. Is that what you had in mind? Thanks!

ekraffmiller:
> @ekraffmiller , filed an issue, and added link in the comment. Is that what you had in mind? Thanks!

Yes, thanks!

ekraffmiller (approved):

PR #218, opened 2025-01-13
Handle public and private CSVs

### LINKED ISSUE #217: Public and private CSV
a user asked for this, and it also aligns with what Stefano would like. The exact wording is not pinned down, but the general idea has buy-in.

mccalluc:
- Fix #217

For the reviewer:
- Does the way the choices are explained in the UI make sense? Coming out of the design meeting, there was feedback that that needed to be improved.
- Some manual testing would be good.
- Will this codebase be maintainable? If this is getting scary, is there anything that should change? Do you think the complexity would be comparable in any other framework?


ekraffmiller:
@mccalluc I think that there are some inherent challenges with Shiny, that are addressed very well in WHAT-WE_LEARNED.md.  I don't know if using another framework would reduce the complexity, but it might make the complexity more manageable. 
Part of the reason we decided on Shiny is that we are hoping using Python will encourage community support and contributions.  Is there any way we can get some feedback to test that theory? Maybe from OpenDP interns?

PR #223, opened 2025-01-23
Capture and show any error messages from notebook execution

### LINKED ISSUE #222: Show details in STDERR if notebook execution fails
We're not capturing the STDERR from the subprocess, so while we get a long stack trace, it doesn't actually give us any useful into. (Hit this when it was trying to read a non-existent file.)

mccalluc:
- Fix #222

Note that [warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn) != [Logger.warning()](https://docs.python.org/3/library/logging.html#logging.Logger.warning). The former is what we're using.

### DIFF
--- a/dp_wizard/utils/converters.py
+++ b/dp_wizard/utils/converters.py
@@ -2,7 +2,7 @@ from pathlib import Path
 from tempfile import TemporaryDirectory
 import subprocess
 import json
-from logging import warning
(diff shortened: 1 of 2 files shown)


ekraffmiller (approved):
Looks good, tests handle the errors as expected

PR #229, opened 2025-02-11
HTML and PDF notebook download

### LINKED ISSUE #55: HTML / RTF / PDF export?
For any of these, probably base it on the notebook and use an off-the-shelf conversion tool?

mccalluc:
- Fix #55
- Rearrange download buttons: Does this look ok?
- PDF was explicitly requested, but HTML probably doesn't hurt?
- Filed a follow-up issue: #230


ekraffmiller (commented):
The downloads work well, but I think the UI for the downloads tab would look nicer with dropdowns for each type (script, report, notebook), rather than clicking the arrow to get extra buttons. What do you think?

Screenshot 2025-02-14 at 7 26 27 AM

mccalluc:
I think you're saying that everything would be in a details drop down? I would really prefer not to require another click to get the notebook: The default choices should just be on the page, and only the secondary options should require another click.

Let me try just putting the other buttons below instead of to the right and we'll see if that looks better.

ekraffmiller:
> I think you're saying that everything would be in a details drop down? I would really prefer not to require another click to get the notebook: The default choices should just be on the page, and only the secondary options should require another click.
> 
> Let me try just putting the other buttons below instead of to the right and we'll see if that looks better.

I was thinking of something like this, and you can make the dropdown prefilled with a default option if you want:


Screenshot 2025-02-14 at 10 57 46 AM

mccalluc:
Cool: Let me review what widgets Shiny provides and see if there's an option that is tidier, but doesn't require too much  specialized html or javascript.

mccalluc:
This isn't exactly what I would have wireframed, but it works with what Shiny gives us. I also realized that giving more information about what each of these is wouldn't be bad. I know all these details, but a user isn't going to know the affordances of each of these.

ekraffmiller (approved):
It looks good! I like the extra descriptions of the different download types.

PR #234, opened 2025-02-12
Cleanup pip output

### LINKED ISSUE #230: Clean up `pip install` outputs cell
Particularly in the PDF, where it causes a page wrap, it's distracting, and in any context it is not very informative.

mccalluc:
- Fix #230


ekraffmiller (commented):
This looks good, the pip install output is cleaned up, but the first page of the PDF is left blank after the install command, which looks a little strange. Is it possible to remove the blank lines after the install command?

mccalluc:
Let me download again: I though that without that output block the wrapping wasn't as bad, but the script block is also large, and it might not be able to break that up either. I'll check.

mccalluc:
PDF render still isn't what I'd like, but I've learned something about the machinery, and if it is a priority, we could come back to this. Unless this is really intolerable, my preference is either to merge this as is, or remove the PDF render entirely. A follow-up issue could be filed.

mccalluc:
Switching from "classic" to "lab" template fixed the page break issue. Also make some other tweaks for readability.

ekraffmiller (approved):
looks good, approved 👍

Issue #219, opened 2025-01-14
Merge `make_accuracy_histogram` into `shared.py`


mccalluc:
We want to make sure the calculations are done the same in both contexts.

PR #237, opened 2025-02-13
Grouping

### LINKED ISSUE #215: Support single group (UI and backend)
We don't have the design finalized, but can still get started (and any roadbumps I hit could influence the design).

mccalluc:
- Fix #215 
- If you apply grouping, the plot is less useful because we don't have room to label all the columns. #247 puts the labels at an angle.
- There isn't an error if you select the same column for both grouping and histogram, although it's not an analysis you would actually want to do.
- We don't use the grouping in the previews: Possible, but would add complexity.
- My inclination is not to put supporting different groupings for different analyses on the road map, and just focus on more analysis options after this. (first up, quantiles)

Suggest checking it out locally and running it before looking at the code.


ekraffmiller (commented):
This looks good, and I think when the new x-axis code is merged, the histogram will be much easier to understand.  I think it's a little confusing to the user though to have the preview histogram not reflect the grouping.  If it's too complex to add that for now,  is there more info we can add to the analysis tab to explain this to the user?

mccalluc:
@ekraffmiller , that's reasonable. I've added a little note for now. The real fix would come with 
- #219

... but I worry that that is something that could get complicated, so I'd prefer not to make that a requirement.

ekraffmiller (approved):
looks good, approving

PR #239, opened 2025-02-13
Pin opendp version

### LINKED ISSUE #236: Pin `pip install 'opendp[polars]'`
The margins interface is going to change in the next OpenDP release, so we need to make sure that the notebooks we build function across that API change: Pin the opendp version when installing.

mccalluc:
- Fix #236
- Move `test_misc.py` up: The test directory structure should reflect the source directory structure, so an extra file like this should just be at the top level.
- Rebuild requirments, since we are upgrading opendp. This also will address the two dependabot warnings.


ekraffmiller (approved):
looks good!

PR #240, opened 2025-02-13
Handle more columns in dataframe helper

### LINKED ISSUE #238: `df_to_columns(histogram_df)` when df has more than two columns
When we support grouping:
- #237

The results dataframe will have more than two columns. Down the road, we might want to carry this richer structure farther downstream, but for now I think it will be easiest to tweak `df_to_columns` to produce new keys which just concatenate the previous ones.

mccalluc:
- Fix #238
- Will help with #237 

The idea is that rather than sorting out right now the details of how visualization will work when we have more keys (ie, should it be multiple plots? Or groups of columns? Or stacked columns? ...), we just concatenate the keys, and everything downstream works the same as before: Just one plain bar chart.

This code is included in the final notebook, and it is getting a little long, so move the test out.

### DIFF
--- a/dp_wizard/utils/shared.py
+++ b/dp_wizard/utils/shared.py
@@ -28,14 +28,23 @@ def df_to_columns(df: DataFrame):
     Transform a Dataframe into a format that is easier to plot,
     parsing the interval strings to sort them as numbers.
     >>> import polars as pl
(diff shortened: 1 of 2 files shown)


ekraffmiller (approved):
looks good, approved

PR #241, opened 2025-02-14
Add a helper for making the changelog, and update the README

### LINKED ISSUE #196: Automate, or at least document, publish process
See what the easiest integration between pypi and github actions is: Don't need anything special.

mccalluc:
The script is a simplified version of what we use on opendp. I haven't run through the steps on the README, but I'll try it when we make v0.2.0.

(OpenDP adds categories in the changelog, so it wouldn't be hard to add, but I don't think it's necessary here.)

(Since it does take the commit message for the changelog, and since the merge commit message is the title of the PR, I should remember to make sure the titles are informative!)

- Fix #196


ekraffmiller (approved):
Looks good, I tried it and the CHANGELOG was updated with commits from the git log.

PR #247, opened 2025-02-24
simplify plotting: No major/minor; instead angle the x-axis labels

### LINKED ISSUE #244: Can x-axis labels be tilted, so we don't need to skip?
Skipping some was fine with numeric values, but with compound strings it becomes a problem.

mccalluc:
- fix #244


ekraffmiller (approved):
Looks good!  Good idea for improving the visualization.

PR #248, opened 2025-02-24
download unexecuted notebooks

### LINKED ISSUE #246: Download un-executed notebooks
Maybe someone wants to share their analysis, but not any data. This would be easy to add. Probably break the download page into two sections: Downloads with results, and downloads that are just the analysis: the later would also include the script download.

mccalluc:
- Fix #246

This isn't a super high priority, but it also seems like an easy addition which could be useful.

There's a lot of redundancy in the function definitions, but reactive calcs don't take parameters, so this seems like an ok approach, but open to suggestions.

### DIFF
--- a/dp_wizard/app/results_panel.py
+++ b/dp_wizard/app/results_panel.py
@@ -103,6 +103,7 @@ def results_ui():
                     """
                 ),
             ),
+            open=False,
         ),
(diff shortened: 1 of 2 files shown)


ekraffmiller (approved):
looks good, approved!

PR #250, opened 2025-02-26
no commas in compound predicates

pdurbin:
An alternative to this PR is to have two independent clauses. Something like this:

"Groups aren't applied to the previews on this page, but they will be used in the final release."

https://www.grammar-monster.com/glossary/compound_predicate.htm


mccalluc (approved):
If you prefer this, sure.

PR #260, opened 2025-02-28
Bump dependency versions and drop 3.9 support

mccalluc:
and also update developer instructions.

We do clear out the nightly builds periodically, but this will probably work until the OpenDP 0.13.0 release, and then we'll update here to match that before the 0.3.0 release.

In the process of this upgrade, filed two issues on opendp. I could be confused, but I think it's good that we have a project that we're developing downstream from opendp that can be a guinea pig for the changes before they hit a wider audience.
- https://github.com/opendp/opendp/issues/2298
- https://github.com/opendp/opendp/issues/2297


ekraffmiller (approved):
looks good, approved

PR #279, opened 2025-03-11
upgrade jinja for dependabot

mccalluc:
not sure why other compiled deps changed. Did this by adding explicit constraint, compiling, removing constraint, and re-compiling, but there should be a better way. This option didn't work for me:
```
  -P, --upgrade-package TEXT      Specify a particular package to upgrade; may
                                  be used more than once
```


ekraffmiller:
@mccalluc I'm not sure I understand your comment...did dependabot create a commit with the required upgrade?  I see that the branch is named dependabot, but I only see your commit on the branch.

mccalluc:
https://github.com/opendp/dp-wizard/security/dependabot/7

Just gave you admin privs, so hopefully you can see this now?

ekraffmiller (approved):

PR #280, opened 2025-03-11
Exercise all downloads

### LINKED ISSUE #274: Parameterize test_default_app_downloads by download type
```
page.get_by_text("Download HTML").first.click()
```
Pull `HTML`, etc., into a parameter.

- Less testing code
- Easier to see the cases we cover
- Might run faster, when we run in parallel?

mccalluc:
- Fix #274


ChengShi-1 (approved):
changes are totally reasonable and make the testing lighter. good to me

PR #281, opened 2025-03-11
Add info to help users compare analysis choices

### LINKED ISSUE #276: UI: Add info to help users compare analysis choices

mccalluc:
- Fix #276

For the reviewer:
- Is the overall layout good?
- Does the text seem like a good explanation of the options? (Recognizing that quantiles haven't been implemented yet.)


ekraffmiller (commented):
The layout looks good, and the text is helpful, I think this is a good addition.  Is there a plan for when quantile will be implemented? I think it would be good to say something in the text about the quantile not being implemented yet, just to avoid confusion.  If quantile is going to be added soon though, I think it's ok as is.

mccalluc:
Thanks for the review. How about if we just keep this on hold until quantiles are implemented? I don't want **TODOs** in the UI, but I also can't give much advice when there are only two options. 

I'll switch this to a draft, and resurface it later.

mccalluc:
Moving this out of draft to indicate that the text is ready, though the median itself is in a separate PR:
- #299

When that is merged to main, merge main to this branch.

ekraffmiller:
> _Calculating a differentially private mean requires that you specify lower and upper bounds so that information outliers is not revealed. If you don't know how the data is distrubuted a histogram or median may be better._

 This seems a little confusing, since all of the analysis options require an upper and lower bound, not just mean.
 


> _With a histogram you can choose a bin count that will answer your question while conserving your privacy budget. If the data is centrally distributed, a mean may suffice, but if that's not the case, a median may be better._
 

 I found this a little confusing as well - it assumes the we know what the user's question is, but maybe we don't?
 Also, it implies that it uses less of the privacy budget than other analysis options, is that true?

ekraffmiller (commented):
looks good, I have some comments about the help text

ekraffmiller (approved):
The updated comments look good, approving

PR #282, opened 2025-03-11
Add an "about" card on the launch page

### LINKED ISSUE #272: Add more info about what the application tries to do inside the app
Either on the launch page or in a modal.

Also update the pypi readme.

mccalluc:
- Fix #272

Stefano requested that there be some kind of description about the app on the launch screen. Exact content and design not specified. Modal was mentioned, but there is plenty of room on this page, so I'd prefer to keep it simple.

For reviewer:
- Design ok?
- Content ok?

### DIFF
--- a/dp_wizard/app/dataset_panel.py
+++ b/dp_wizard/app/dataset_panel.py
@@ -32,51 +32,83 @@ def dataset_ui():
 
     return ui.nav_panel(
         "Select Dataset",
-        ui.card(
-            ui.card_header("Input CSVs"),
-            ui.markdown(
-                f"""
+        ui.layout_columns(
+            [
+                ui.card(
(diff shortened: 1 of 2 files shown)


ekraffmiller (commented):
The card on the right looks good, but the info is not accessible after the continues to the 'Define Analysis' step.  Another way to do it might be to have the info in a tab, which is displayed first, and that the user can go back to.  It could be called 'About', which is a common naming convention for this type of information. Also having the info in its own tab means there would be more space if you want to provide additional content.  Usually 'About' is a link, but it seems like a tab is easier to add to the current implementation, so thats why I suggested it. I think either would be ok. What do you think?

For the content, maybe a high level description of differential privacy would be good? For example, DPCreator has this introductory text -  

Image

PR #283, opened 2025-03-11
pyyaml does not need to be installed in notebook

### LINKED ISSUE #268: pyyaml install can be moved into coda of notebook
None of the exported user code references it, so we could move it into the coda

mccalluc:
- Fix #268


ekraffmiller (approved):
works fine without the dependency, approved

PR #284, opened 2025-03-11
Only access CLIInfo from server

### LINKED ISSUE #273: Inconsistent access to CLI info
It doesn't look like the end-to-end demo test actually exercises `--demo`:

In the tests:
```
app = App(
    app_ui,
    make_server_from_cli_info(
        CLIInfo(
            public_csv_path=None,
            private_csv_path=None,
            contributions=1,
            is_demo=True,
        )
    ),
)
```
```
def make_server_from_cli_info(cli_info: CLIInfo):
    def server(input: Inputs, output: Outputs, session: Session):  # pragma: no cover
        public_csv_path = reactive.value(  # noqa: F841 # TODO
            cli_info.public_csv_path or ""
        )
        private_csv_path = reactive.value(cli_info.private_csv_path or "")

        contributions = reactive.value(cli_info.contributions)
```

In contrast, in `dataset_ui`, the `CLIInfo` is coming from the environment, and it never sees our manual override:
```
def dataset_ui():
    cli_info = get_cli_info()
```
```
def get_cli_info() -> CLIInfo:  # pragma: no cover
    args = _get_args()
    if args.demo:
        return _get_demo_cli_info()
```
```
def _get_args():
    """
    >>> _get_args()
    Namespace(public_csv_path=None, private_csv_path=None, contributions=1, demo=False)
    """
    arg_parser = _get_arg_parser()

    if "pytest" in argv[0] or ("shiny" in argv[0] and "run" == argv[1]):
        # We are running a test,
        # and ARGV is polluted, so override:
        args = arg_parser.parse_args([])  # pragma: no cover
    else:
        # Normal parsing:
        args = arg_parser.parse_args()  # pragma: no cover
```

My guess is that we should get rid of all the special logic in UI code, and make sure the information flow is always from the server.

mccalluc:
- Fix #273

... I think: This is complicated, I could have missed something.

### DIFF
--- a/dp_wizard/app/__init__.py
+++ b/dp_wizard/app/__init__.py
@@ -25,9 +25,7 @@ def ctrl_c_reminder():  # pragma: no cover
 
 def make_server_from_cli_info(cli_info: CLIInfo):
     def server(input: Inputs, output: Outputs, session: Session):  # pragma: no cover
-        public_csv_path = reactive.value(  # noqa: F841 # TODO
-            cli_info.public_csv_path or ""
-        )
+        public_csv_path = reactive.value(cli_info.public_csv_path or "")
         private_csv_path = reactive.value(cli_info.private_csv_path or "")
 
         contributions = reactive.value(cli_info.contributions)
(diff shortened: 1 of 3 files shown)


ChengShi-1 (approved):
inspected codes, UI works well, and run test successfully. It looks good to me.

PR #287, opened 2025-03-12
Add "about" tab

### LINKED ISSUE #272: Add more info about what the application tries to do inside the app

mccalluc:
- Replaces #282
- Fix #272

I would like to keep this short, and if there is important information about what this application does, then it really belongs in a README: Users shouldn't need to install to get this information.

The situation is different from DP Creator. This, at present, is a single user application, and the user found and installed it themselves. It is not embedded in a larger application.

I would strongly prefer that most users not be slowed down by this. If there's a request that this be on a splash screen on the first run, I'd prefer to tackle that as a follow-up.

About pages often do include version numbers, particularly to help users filing issues, so I've included that here as well.

### DIFF
--- a/README-PYPI.md
+++ b/README-PYPI.md
@@ -1,7 +1,7 @@
 # DP Wizard
 
 DP Wizard guides the user through the application of differential privacy.
-After selecting a local CSV, users are prompted to describe to the anlysis they need.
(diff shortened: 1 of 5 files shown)


ekraffmiller (approved):
Looks good! approved

Issue #257, opened 2025-02-27
Select too many columns and it just breaks


mccalluc:
Construct a CSV with 20 columns and just keep adding them. Shiny can't keep up and the UI becomes unresponsive.

I think (but haven't tested this hypothesis) that plot generation is expensive.
- Display only one preview plot at a time?
- Hard limit on the number of columns that can be selected?
- Less recalculation? Make things lazier / less reactive?
- More caching of results?

mccalluc:
Started a PR
- #289

but I wonder if part of the solution is [non-blocking operations with Shiny](https://shiny.posit.co/py/docs/nonblocking.html). Need to do more research.

mccalluc:
There are at least two problems using non blocking operations. 😦 

They are intended as a wrapper for calculations: What was reactive is now called as a regular function. Since the slow part for us is the plot, I'm not sure how to change this to a function invocation:
```
ui.output_plot("histogram_preview_plot", height="300px"),
```

But even if that worked, while non blocking operations don't block _other_ code, they [still block themselves](https://shiny.posit.co/py/docs/nonblocking.html#multiple-invocations):
> An extended task can run concurrently to reactive code and to other extended tasks–that’s its whole purpose. However, a single extended task object cannot run itself multiple times concurrently. If you call `sum_values()` while it is already running, it will enqueue the new invocation and run it after the first one completes. 

They acknowledge that this is often not desired: The fix is a widget that blocks the user from enqueing more operations. I don't know that a UI that blocked user operations would be any better than one where they get backlogged.

👉 New idea: There are many plotting libraries out there: speed might not usually be the deciding factor, but worth looking at in this case.

mccalluc:
No obvious switch that just makes it faster. possible next steps
- Time the rendering code: if we'll be making experiments to speed this up, we need numbers. Playing with the UI and seeing how it feels is not enough.
- Switch out [matplotlib backends](https://matplotlib.org/stable/users/explain/figure/backends.html). Does it make a difference?
- Profile the whole thing: I've assumed `plot_histogram` is the bottleneck, but not super confident
- Is it incrementally updating the plot? That could be expensive. Make sure we're only rendering once.

Issue #256, opened 2025-02-27
Allow column removal from lower in the page


mccalluc:
> Allow users to remove selected columns directly within the block instead of scrolling up to uncheck them—adding a "Remove" or "Delete" button would  be good.

PR #289, opened 2025-03-13
Add a warning about a large number of columns

mccalluc:
- Toward #257

There might be changes at a lower level that could improve performance, but after giving it a morning, I'm not seeing any easy wins. (More notes on issue #257.) This would add a warning message if the user adds five columns, which is when performance seems to be really impacted.

- The warning itself is obscured by the selectize list; This isn't great, but I don't think there's a better place to put it: We want it near by, and I don't think we want it above. (Generally a bad idea if elements of the UI are moving around.)
- It's not a hard stop. If #256 is implemented, and there is another way to remove columns, we could disable this control when too many columns are selected... but until/if that's implemented, we don't want to disable this control, because it's the only way to remove columns.

So take this as a proposal: I'd be ok with closing this PR if it doesn't seem better than the status quo.

### DIFF
--- a/dp_wizard/app/analysis_panel.py
+++ b/dp_wizard/app/analysis_panel.py
@@ -11,7 +11,7 @@ from dp_wizard.utils.csv_helper import (
     read_csv_ids_names,
     get_csv_row_count,
 )
-from dp_wizard.app.components.outputs import output_code_sample, demo_tooltip
+from dp_wizard.app.components.outputs import output_code_sample, demo_tooltip, info_box
 from dp_wizard.utils.code_generators import make_privacy_loss_block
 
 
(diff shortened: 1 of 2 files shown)


ChengShi-1 (commented):
I just tried the UI. The warning appears only after the data graph has fully loaded. I think it would be good if the warning is shown immediately once the user selects >=5 columns, without waiting for the data graph to load. This way, users will know what to expect while waiting.

mccalluc:
@ChengShi-1 : The framework restricts events to being handled in the order they are received. So when I add one column, and another, and then a third, it needs to handle all of the updates for those before it can do anything else, including showing an error message.

I don't have any ideas for how to improve this right now. Happy to keep the underlying issue open, in case a better solution occurs to any of us.

Closing.

Issue #278, opened 2025-03-10
Make histogram vs mean OO


mccalluc:
We have a whole lot of parallel case statements. These would be better with OO. This would also move us towards "analysis modules".

mccalluc:
I'm going to consider this closed: Code generation has been refactored. Looked into doing something similar for the UI code, but got stuck, in part because there's a lot more shared between the different analyses, and in part because refactoring Shiny modules is just harder, and when something is wrong there's not much feedback.

PR #290, opened 2025-03-14
Make Analysis templates OO

mccalluc:
- Towards #278 

Keep the issue open because I would also like to handle the UI in here, but that's a little trickier because we need access to the reactive context. Probably use a shiny module?

The goal of this is to make the maintenance and creation of new analysis options easier by bringing together everything that is needed for a particular analysis in one place.

Parallel type checking across a codebase is a good sign that OO would be a better approach, and that's what we had before this PR. Didn't end up actually using `class` because what I really wanted is the singleton pattern, and adding explicit class syntax, and then treating those as singletons, doesn't buy us much. The downside of this approach is that we don't have anything like an abstract superclass with all the abstract methods we need to make this concrete, but I think that's ok: If something is missing, the errors will make it pretty obvious.

For the reviewer to consider:
- If you were tasked with adding a new analysis option, does this seem easier than the old way?


ekraffmiller (commented):
Hi Chuck, I think this looks good, I like the organization, and it makes it more clear what needs to be included when adding a new Analysis.  I know it's a matter of opinion, but I think it would be nice to have something in the code that defines the required functions in the analysis modules, even if they are meant to be singletons.  
Have you considered using Python Protocol to define the interface?  That might be a way to do it without a lot of overhead.

mccalluc:
How about this? Ran into circular import problems, but the protocol does make the desired shape more explicit. (Though even without this, pyright has been smart enough to know what could be returned, and gives warnings if a needed function is missing.)

mccalluc:
I also spent some time trying to apply a similar refactor to the UI code, and didn't finish the work. Challenges.
- The shiny module system is just hard to work with, and when a required method is missing there's not much feedback.
- On a first pass, I tried to factor out the inputs, but then the `reactive.event` handlers also need to be duplicated, and that leads to a lot of duplication, and risks leaving out critical parts.
- On a second pass, I tried pulling out just the preview code... but there are references to the inputs which are then unsatisfied.

ekraffmiller:
> How about this? Ran into circular import problems, but the protocol does make the desired shape more explicit. (Though even without this, pyright has been smart enough to know what could be returned, and gives warnings if a needed function is missing.)

I think this looks good! I think having the type more explicit will be helpful for code readability.

ekraffmiller (approved):
Looks good, the OO design helps with the organization of the analysis code.  👍

PR #291, opened 2025-03-17
Move UI code into OO

mccalluc:


mccalluc:
Refactoring the UI code is hard, and even if it did work, it's not clear that adding more abstraction there would make it easier to maintain, which is the ultimate goal.

Closing.

Issue #249, opened 2025-02-26
tool tip way on the right


pdurbin:
In on bd33876 and this tool tip seems to be placed in a funny location to me, way on the right:

![Image](https://github.com/user-attachments/assets/bf76a78b-7acb-4b73-b94a-715a2bb1a7c6)

pdurbin:
Another example but less extreme since the card only takes up half the screen, horizontally:

![Image](https://github.com/user-attachments/assets/7bc3317f-0da5-4e83-9d2a-bce6d54c999b)

mccalluc:
I fought with the CSS for a little while, but since I'm not confident that the tool-tips are the right approach didn't want to sink too much time into this. 

Maybe the demo notes should be visible by default, instead of hidden away? I like the way the demo mode doesn't depart too far from the non-demo, but maybe that's not the thing to optimize. This might be a good thing to a wider range of feedback on.

PR #292, opened 2025-03-18
Hacky CSS for a better epsilon slider

### LINKED ISSUE #243: Fix rendering of epsilon slider
As originally implemented, it looked fine, but now that it's contained in a card it is either scrunched, or the label for "10" is floating on the right.

First step is probably to see how the vanilla control works, before looking at our customizations.

mccalluc:
- Fix #243

### DIFF
--- a/dp_wizard/analyses/protocol.py
+++ b/dp_wizard/analyses/protocol.py
@@ -1,0 +1,43 @@
+from typing import Protocol
+
+from dp_wizard.utils.code_generators import CodeGenerator
+
+
+class Analysis(Protocol):  # pragma: no cover
+    # There should also be a "name".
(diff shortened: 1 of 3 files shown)


ChengShi-1 (commented):
It looks nice! Do you think we could make the tooltip closer to the slider? It seems a bit distant right now. Or, you could put it after the number `Epsilon: 0.794` below the bar
image

mccalluc:
That isn't a regression with this PR, and except from being adjacent, it is unrelated to the slider. There is a separate issue:
- #249

I'd like to address that too, just not in this PR. Can you re-review with that in mind? Thanks!

ChengShi-1 (approved):
Looks good to me

PR #293, opened 2025-03-18
Fill in codegen gaps for means

mccalluc:


dp_wizard/utils/shared.py
```diff
@@ -20,8 +20,10 @@ def interval_bottom(interval: str):
     """
     >>> interval_bottom("(10, 20]")
     10.0
+    >>> interval_bottom("-10")
+    -10.0
     """
-    return float(interval.split(",")[0][1:])
+    return float(interval.split(",")[0].replace("(", ""))
```
ChengShi-1, line 26: There might be cases like `interval_bottom("[5, 15)")`, which wouldn’t be handled correctly by just using replace("(", ""). If that’s not relevant to the data, feel free to ignore this comment.

dp_wizard/analyses/mean/no-tests/_mean_script_output.py
```diff
@@ -1 +1,9 @@
-# todo: https://github.com/opendp/dp-wizard/issues/275
+print(
+    f"DP means for COLUMN_NAME "
```
ChengShi-1, line 2: COLUMN_NAME here may need to be dynamic?

ChengShi-1 (commented):
Good! just notice something small

dp_wizard/utils/shared.py
mccalluc: The intervals like `(0, 10]` come from Polars after applying cut points, and are always open on the left and closed on the right (except for `(..., inf)`). But I can add a comment here to explain the particular case we want to catch.

dp_wizard/analyses/mean/no-tests/_mean_script_output.py
mccalluc: It is. All of these template files substitute values where there are string in ALLCAPS.

mccalluc: ... but if you're just looking at this file in isolation, how would a new developer know that? There's another PR
- #301

that would let us pull these templates closer to where they are used, and that might help.

mccalluc:
One little comment added about intervals. The templating could also use a better explanation, but there might be bigger changes there soon-ish. Thanks for review, @ChengShi-1.

ChengShi-1 (approved):
Looks good to me

PR #295, opened 2025-03-19
Remove columns from lower on page

### LINKED ISSUE #256: Allow column removal from lower in the page

mccalluc:
- Goal: Fix #256

but not working right now. Will file as draft and come back to it at some point.


mccalluc:
Still a good idea, but doubt that keeping this PR around moves us closer to it. Closing.

Issue #262, opened 2025-03-03
With delta, margins can be dropped


mccalluc:
Suggestion from Mike.
- Use `delta` (already in place)
- Add an explanation of delta, and what it gets us
- Remove `margins` kwarg
- move `with_columns` out of preprocessing, and down between `query()` and `group_by()`

PR #299, opened 2025-03-20
Support median

mccalluc:
Tests currently failing, but UI worked?


mccalluc:
Filed an issue upstream:
- #2331

Will wait for feedback on that before proceeding here.

ekraffmiller:
> 
> * Up the minimum contributions to 2 temporarily to avoid [calculating median: `contributions=1` + `public_info="lengths"` causes `unable to infer bounds` opendp#2331](https://github.com/opendp/opendp/issues/2331) -- but the **real fix** is [With delta, margins can be dropped #262](https://github.com/opendp/dp-wizard/issues/262)
>
@mccalluc I tried this out, and I'm still allowed set Contributions = 1.  Should it be validating to prevent contributions < 2?

mccalluc:
> > * Up the minimum contributions to 2 temporarily to avoid [calculating median: `contributions=1` + `public_info="lengths"` causes `unable to infer bounds` opendp#2331](https://github.com/opendp/opendp/issues/2331) -- but the **real fix** is [With delta, margins can be dropped #262](https://github.com/opendp/dp-wizard/issues/262)
> 
> @mccalluc I tried this out, and I'm still allowed set Contributions = 1. Should it be validating to prevent contributions < 2?

Sorry not to update that! I used it as a work around earlier in the history of this PR, but the last commit (`lengths -> keys`) removes that constraint, but I forgot to update the text.

ekraffmiller (approved):
looks good, I how the Analysis code is organized 👍

PR #301, opened 2025-03-21
Use functions as templates

### LINKED ISSUE #300: Refactor code templating: function rather than string?

mccalluc:
- Fix #300

Working with the templates, it has gotten a little confusing switching between the template file, and the definition that fills the template. The idea is that we could use functions in the main source code for this, using the `inspect` module to pull out the source code lines.

This PR just introduces the machinery and tests: a follow-up would get rid of the small, standalone template files... if this even seems like a good idea.

Reviewer:
- Review the code itself
- Feedback on the idea: will these template functions be easier to read than code in separate small files?


ekraffmiller (commented):
I don't see immediately how this will improve the usage of the template code, but maybe because I'm just seeing the small examples in the test code.  Is the main difference that the function version of the Template will include a list of parameters, so it's easier to see what strings are needed to fill the template, rather than looking though the file for all caps terms?

ekraffmiller:
Ok, that makes sense thanks. I can see how having the small template files incorporated in this way can make the code easier to manage.

ekraffmiller:
I'm not sure if I should approve this since it will do an auto-merge. Do you want this merged to develop right away, or is it still experimental?

mccalluc:
> I'm not sure if I should approve this since it will do an auto-merge.

Huh: When I look at it, it doesn't seem to be enabled... but even if it were, that would be fine. I would like to get the machinery in now, and down the road we can actually use it.

ekraffmiller (approved):
approved

PR #302, opened 2025-03-21
Fix tool tips that have been floating to the right

### LINKED ISSUE #249: tool tip way on the right

mccalluc:
- Fix #249


ChengShi-1 (approved):
Looks good to me. It works perfect on my end

PR #303, opened 2025-03-21
A little bit of input validation

### LINKED ISSUE #254: Input validation: lower < upper

mccalluc:
- Fix #254

but there is more that can be done!


dp_wizard/app/components/column_module.py
```diff
@@ -19,6 +24,31 @@
...
+    >>> get_float_error('0')
+    >>> get_float_error('')
+    'is required'
+    >>> get_float_error('1.1')
+    >>> get_float_error('nan')
+    'should be a number'
+    >>> get_float_error('inf')
+    'should be a number'
+    """
+    if number_str == "":
```
ChengShi-1, line 38: I don't see it has `... is required`, even I don't have any value inside the input box, this condition seems to be covered by following conditions.
image

ChengShi-1 (commented):
I created an issue for a bug I found about the number stepper https://github.com/opendp/dp-wizard/issues/306

dp_wizard/app/components/column_module.py
mccalluc: This now warns about the missing value, but also emits the same error if you provide a string that can't be parsed to a number (like "abcxyz"). I think Shiny tries to convert to a number before we even see it, so those messages can't be reached... but if 
- #310

were merged, the strings would be visible to validation functions, so let's keep it around, even if it's not used in this branch right now.

ChengShi-1 (approved):
Good, approving

Issue #306, opened 2025-03-24
number steppers aren’t behaving correctly


ChengShi-1:
Please provide relevant environment details, a minimal reproducer for the bug, and the expected behavior.

When I click the number stepper to +1 or -1 multiple times at once, it keeps jumping

https://github.com/user-attachments/assets/183db741-7795-458c-aefb-3a7fc0fb02c3

mccalluc:
To reproduce, you need to increment twice, quickly: before the the UI has caught up.

This bug was introduced with
- #264 

With the conditional rendering of the controls, my guess is that in some sense there are two instances of these controls, one for histogram and one for mean, and they are fighting over what the current value is. Not sure what the fix will look like, but I'll get started.

PR #310, opened 2025-03-26
`text_input` + `update_on="blur"` reduces unnecessary re-renders

mccalluc:
Started this to address
- #306 

but doesn't actually fix that: two changes in quick succession could still cause a loop.

But: I think this still may be an improved experience. If you're entering a number, it's not helpful to show a result for "1", and then "12", and then "123": Better to just wait until input is complete.

(`update_on="blur"` requires upgrading shiny.)

This is subjective. What do you think?


ekraffmiller:
@mccalluc I tried it out - I think it's good to remove the incremental arrows from the input, and using onBlur prevents re-renders.  I think it seems a little awkward though to have to click outside the input to trigger rendering.  
Another solution would be to implement debouncing - to introduce a small delay (usually about .3 seconds) before triggering the re-render.  What do you think?

mccalluc:
> Another solution would be to implement debouncing - to introduce a small delay (usually about .3 seconds) before triggering the re-render. What do you think?

My initial though was that we don't want to get into too much widget development if we're using Shiny. (It's a library that keeps things simple... so we need to keep things simple.)

**But** it looks like there's a [gist that does exactly this](https://gist.github.com/jcheng5/427de09573816c4ce3a8c6ec1839e7c0).

> Why is this a gist instead of a PR? Because to implement this "properly" for Shiny for Python, this would need to be beefed up to include type annotations, support for async, and unit/integration tests, which would all be more effort than this has taken so far. (To be clear, we want to do all that, but in the meantime, here's this gist.)

(linked from https://github.com/posit-dev/py-shiny/issues/564)

I'll try that out on this PR and ping you for re-review. Thanks.

mccalluc:
Second thought: 
- text input feels like it's an easy improvement. (Keeping it on contributions, because that's a place where we actually do want small integers.)
- but debounce and/or blur require more changes.

Simplest way forward is probably to make a new PR with just the change to text input, and save this PR for experiments.

mccalluc:
Closing in favor of 
- #335
- #336

PR #311, opened 2025-03-26
reactive isolate fixes infinite loop

### LINKED ISSUE #306: number steppers aren’t behaving correctly

mccalluc:
- Fix #306

If we just see a change in the reactive values, we _should not_ redraw the inputs.

(Turn on "ignore whitespace in diff" in github to make the review easier).

I went down a little dead end before remembering this method: It's not clear to me that the increment/decrement buttons are actually a good thing, or that the plots should rerender on every keypress. A separate PR addresses that, but doesn't really fix the bug:

- #310

### DIFF
--- a/dp_wizard/app/components/column_module.py
+++ b/dp_wizard/app/components/column_module.py
@@ -131,51 +131,53 @@ def column_server(
         }
         match input.analysis_type():
             case histogram.name:
-                return ui.layout_columns(
-                    [
-                        ui.input_numeric(
-                            "lower_bound",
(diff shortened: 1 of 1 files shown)


ekraffmiller:
Looks good, using the increment/decrement updates the visualization without causing an infinite loop

ekraffmiller (approved):
approved

PR #312, opened 2025-03-26
No silent errors for code gen

### LINKED ISSUE #305: If notebook build fails, show something in the UI
If it's possible to pull a little info from the log, great, but a vanilla message would be fine.

mccalluc:
- Fix #305

Code review is sufficient, but if you'd like to see the modal, try adding an error at the top of `_notebook.py`. 

Shiny doesn't offer component level testing, and since the app if it's working correctly never hits this code, it's not easy to execute in an end-to-end test. If we need to do that, the solution might be adding functionality which is hidden from the user, or making a mini application just to exercise this.

I don't think that's necessary right now, but that might not be the right call. Could also file an issue and come back to this later.

### DIFF
--- a/dp_wizard/app/results_panel.py
+++ b/dp_wizard/app/results_panel.py
@@ -108,7 +108,7 @@ def results_ui():
     )
 
 
-def generate_code_or_modal_error(download_generator) -> str:
+def make_download_or_modal_error(download_generator) -> str:  # pragma: no cover
     try:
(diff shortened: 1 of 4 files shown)


ekraffmiller (approved):
I tested it locally and it works fine.  I don't think its necessary now to add a test for it, but we could make it an issue to work on later.

PR #327, opened 2025-04-01
Prompt for lower and upper bounds on dataset size

### LINKED ISSUE #324: `max_partition_length` and `delta` should connect to user inputs
We want to be setting a good example, and just choosing a really big number and a really small number is not that! This should come from the user. One possibility:

- Retitle the "Simulation" section to "Size"
- Change the text, and add an "Estimated Upper Bound" and "Estimated Lower Bound" on the size.
- Use "Estimated Upper Bound" for `max_partition_length`
- User 1/"Estimated Lower Bound" for `delta`

mccalluc:
- Fix #324


mccalluc:
- Taking a different direction on this: #559

PR #337, opened 2025-04-02
provide command line alias

### LINKED ISSUE #331: on commandline make sure `dp_wizard` and `dp-wizard` both work
I think there are good reasons to use a dash in some contexts and and underscore in others (and just a space in plain text), but can we make sure there is an alias set up after the pip install?

mccalluc:
- Fix #331


ekraffmiller (approved):
approved

PR #345, opened 2025-04-03
Add more links to docs

### LINKED ISSUE #340: More links to opendp docs in generated notebook
At least one link from each analysis section.

mccalluc:
- Fix #340

For the reviewer:
- Does this seem like enough? Should we add more?
- Ok that these are comments in the code? We could insert a markdown block just before this code block... but that would be a bit of new programming, for a relatively small improvement in functionality (ie, being able to click on the links in the markdown.)


ekraffmiller (approved):
I think it looks good, helpful to have a link to the related page in the OpenDP docs. I think it's ok as is, without the Markdown.

PR #348, opened 2025-04-03
Move help to the private csv selector

### LINKED ISSUE #332: Demo: public data info
Explain that it can be left blank, and move the existing pop-up over to private.

mccalluc:
- Fix #332


ChengShi-1 (approved):
It's shown correctly on the page. Looks good

PR #349, opened 2025-04-03
Capture environment info, and prepopulate github issue

### LINKED ISSUE #319: Include git info on the about page?
Even distinguishing whether this is a git checkout or a pip install would be useful.

### LINKED ISSUE #320: Include ARGV info on the about tab
Behavior is different with and without `--demo`, and there might be other changes in the future. Include any command line arguments on the about tab.

### LINKED ISSUE #321: Include `pip freeze` on the about tab
It would be a lot of lines, but the information can useful for reproducing errors. Maybe put this and all the other debugging info inside a text area?

mccalluc:
- Fix #321
- Fix #320 
- Fix #319 
- Link to pre-populated github issue

Started out just planning to support ARGV, but the others were easy, and then with all the info in one place, making it easy to file an issue seemed smart... but it could be argued that it's not worth the added complexity.

I first did it as an `<a>` element, and handled the construction of the url myself, and then thought it might be simpler to use a form... but we need JS to override the shiny defaults, so I'm not sure it was simpler in the end. Feedback welcome!

Long term, if we don't have many people submitting the google form, I'd like to drop that, and just focus on this as the best way to share feedback, since it gives us more information.


dp_wizard/app/about_panel.py
```diff
@@ -22,8 +82,19 @@ def about_ui():
...
             ),
-            ui.p(f"DP Wizard version {version}"),
-            ui.p(f"Python {sys.version}"),
+            tags.textarea(
+                info,
+                readonly=True,
+                rows=10,
+                style="font-family: monospace;",
+            ),
+            ui.a(
```
ChengShi-1, line 91: when i click the File Issue button, it didn't jump to the target url directly, but I could go to by right clicking -> "Open Link in the New Tab"

ChengShi-1 (commented):
I'm not sure about our users, if most of them are developers, filing an issue on GitHub could be a good option. Otherwise, maintaining a form might be easier for them. 

As for the pre-populated information, we could include a fold/unfold toggle and keep it collapsed by default, since users probably won't care much about those details. They're more helpful for us during debugging (from my understanding) and showing them upfront might be confusing.

dp_wizard/app/about_panel.py
mccalluc: Huh: It is using `target="blank"`, and maybe your browser prevents new tabs from being opened? Could you see if there is a warning in the browser console? Also, what browser are you using?

mccalluc: > As for the pre-populated information, we could include a fold/unfold toggle and keep it collapsed by default, since users probably won't care much about those details. They're more helpful for us during debugging (from my understanding) and showing them upfront might be confusing.

Maybe there's a better way to present this, but if people are filing issues, I want as much information as possible, and if we require an extra step to see the details, we won't get the details. If we can get the button to work, great, but if we can't, then we really need to make it easy for the user. The `pip freeze` output is long, but it is also particularly valuable for debugging. (Perhaps less important if we can pin all dependencies: See #370.)

ChengShi-1: > Huh: It is using `target="blank"`, and maybe your browser prevents new tabs from being opened? Could you see if there is a warning in the browser console? Also, what browser are you using?

There is no warning in the console, I am using google chrome. It works after removing `class_="action-button"` since there is a conflict `ui.a` and `class_="action-button" `

From what chatgpt said, 
> When you add the class_="action-button" manually to a link (< a > tag), the system starts treating it like a trigger — but that breaks normal browser behavior (like opening in a new tab), because now the framework wants to "observe" it rather than just let the browser follow the link.

mccalluc:
Thanks for feedback. Could you check if it works for you now? I also added another line in the generated issue.

ChengShi-1 (approved):
works good

PR #355, opened 2025-04-04
Add a warning on the first run

### LINKED ISSUE #351: Add a note in the log that first run will be particularly slow
Are there any files being cached that we should look for, so make this conditional?

mccalluc:
- Fix #351

For the reviewer: Is this visible, without being obtrusive?

### DIFF
--- a/dp_wizard/__init__.py
+++ b/dp_wizard/__init__.py
@@ -1,6 +1,7 @@
 """DP Wizard makes it easier to get started with Differential Privacy."""
 
 from pathlib import Path
+from logging import info
 
 
 __version__ = (Path(__file__).parent / "VERSION").read_text().strip()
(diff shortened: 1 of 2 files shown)


ChengShi-1:
Hey Chuck, I just tried, but not sure how to get this to show up on the UI

mccalluc:
(This will be in the console, not the web page.)

ChengShi-1:
Thanks Chuck, I think `WARNING:root:First startup may take a minute; successive runs will be faster.` right now is fine, just be a little bit flooded.

> WARNING:root: ====== First startup may take a minute. Successive runs will be faster. ======

This one could be noticed enough for me.

Also, it could be this as well

> WARNING:root: -------------------------------
WARNING:root: | First startup may take a   |
WARNING:root: | minute. Successive runs    |
WARNING:root: | will be faster.                |
WARNING:root: -------------------------------

ChengShi-1 (approved):
Looks great

PR #358, opened 2025-04-04
minimum version on pyarrow

### LINKED ISSUE #357: `Invalid or unsupported format string: 'vu'`
- Root cause: https://github.com/pola-rs/pyo3-polars/issues/66

The error is coming from pyarrow , which is an unpinned transitive dependency. I have pyarrow==19.0.1, but Ceilyn's environment already had pyarrow==11.0.0, and nothing said a more recent version was necessary, and it broke.

- Find a lower bound on the acceptable version
- pin it
- is it possible to pip install with the oldest versions of all dependencies?

mccalluc:
- Fix #357

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,9 @@ dependencies = [
     "ipykernel",
     "black",
     "pyyaml",
+
+    # Pin minimum versions if we've had problems.
+    "pyarrow>=16.0.0",
 ]
 
 [project.scripts]


pyproject.toml
```diff
@@ -20,6 +20,9 @@ dependencies = [
     "ipykernel",
     "black",
     "pyyaml",
+
+    # Pin minimum versions if we've had problems.
```
ekraffmiller, line 24: would it be better to pin this to an exact version? That way we control exactly what is being used in the app.

mccalluc: That is possible, but there are downsides.
- Pinning to exact version numbers makes it much harder to even think of integrating this tool into a larger project.
- and makes it harder to do upgrades when they are necessary.
- This is a transitive dependency... even if we had pinned the top level dependencies, it wouldn't necessarily have caught this.
- I have filed #361

This is also partially a cultural question. No single dependency manager has a "lock" on the market in python. I think what I'm doing lines up with common Python practices, but I could be wrong. If we do want to handle this differently, I'd look for alternatives withing the python ecosystem.

ekraffmiller: Ok, I'm basing my thought on what is done in package.json for javascript, we always use exact versions. But if it's a different convention in Python,  that's ok

ekraffmiller (approved):

PR #366, opened 2025-04-08
Set lower bound on dependencies

### LINKED ISSUE #361: Set lower bounds on top level dependencies
Ceilyn had an older version of a transitive dependency installed, and that caused problems.
- #357

It should be feasible to set lower bounds on all of our top-level dependencies. While this isn't an absolute guarantee that install will work, it would make this kind of problem much less likely.

Cursory search didn't find a project that automatically searched the space of versions, but even doing it by hand doesn't seem infeasible.

At the end, should there also be a requirments.txt pinned to these minimum requirements, so we know they work going forward?

mccalluc:
- Fix #361 
- prompted by #357

This is not the lowest possible bound, but it does lessen the chance of picking up an old-old-old version of a dependency from the environment.

See also:
- #356 (gives us some coverage of the low and high end of dependency versions, as well as the low and high python versions)
- #60

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,16 +10,16 @@ license = {file = "LICENSE"}
 classifiers = ["License :: OSI Approved :: MIT License"]
 dynamic = ["version", "description"]
 dependencies = [
-    "shiny",
-    "faicons",
-    "matplotlib",
+    "shiny>=1.2.1",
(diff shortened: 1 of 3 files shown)


mccalluc:
What's here isn't bad, but let me see if I can make this stronger.

- #370

Moving to draft.

PR #372, opened 2025-04-10
Link to epsilon registry

### LINKED ISSUE #365: Link to registry
We can't explain all that might go into picking epsilon, but we could link to https://registry.oblivious.com/

mccalluc:
- Fix #365

... but is this even a good idea? For this, and for a lot of other issues, I don't think the full design review process that we had at the start of the project is necessary, but it's also possible that this isn't the right direction. Maybe the emphasis should be on explaining the epsilon choice in this application, rather than foisting that responsibility on to another project?

### DIFF
--- a/dp_wizard/app/analysis_panel.py
+++ b/dp_wizard/app/analysis_panel.py
@@ -55,6 +55,9 @@ def analysis_ui():
                     What is your privacy budget for this release?
                     Values above 1 will add less noise to the data,
                     but have a greater risk of revealing individual data.
+                    Consider how your budget compares to that of
+                    <a href="https://registry.oblivious.com/#public-dp"
+                       target="_blank">other projects</a>.
                     """
                 ),
                 log_slider("log_epsilon_slider", 0.1, 10.0),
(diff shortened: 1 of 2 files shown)


ChengShi-1 (commented):
The public registry could be included as a reference, but I've noticed that the range of epsilon values varies widely from project to project. Since it seems that the e really depends on the data and use case, I'm not sure how much it would actually help users choose an appropriate value.

Also, I see there's already a tooltip explaining the downside of using an epsilon greater than 1. Do we still need to repeat the message _Values above 1 will add less noise to the data, but have a greater risk of revealing individual data_? It might be redundant.

mccalluc:
How about if I put this on the agenda for the next meeting? We have a high level goal about helping people use DP, and part of that might be connecting them to resources... but if it's not actually helpful, what then?

mccalluc:
Talking with Ceilyn and Ellen: Expand the text, and explain that the daily interval and other concerns.

ChengShi-1 (approved):

PR #373, opened 2025-04-10
config -> expr

### LINKED ISSUE #316: `*_config` to `*_expr`
Consistency is good. `config` is also a very general term: `expr` is probably better, but get confirmation.

mccalluc:
- Fix #316

Calling it "config" to begin with was a mistake on my part. These are [polars expressions](https://docs.pola.rs/api/python/dev/reference/expressions/index.html), and `expr` is used consistently in our documentation.


ChengShi-1 (approved):
Looks good, and I agree it seems better to have `expr`

PR #379, opened 2025-04-10
nav button on right

### LINKED ISSUE #378: Nav buttons: align right, and add an arrow

mccalluc:
- Fix #378

There are lots of choices for arrows, either plain characters or font awesome. Let me know if you think something else would look better. (Bootstrap v5 is built-in with shiny, so that's where `float-end` comes from.


ChengShi-1 (approved):
Looks good and works well on my local page

PR #381, opened 2025-04-10
click outside error modal or hit esc to close

### LINKED ISSUE #380: Error modal closes when you click outside

mccalluc:
- Fix #380

(If you want to actually exercise this, you'll need to introduce an error in the notebook, like a `0/0` at the top of `_notebook.py`, though I'm not sure you need to be that diligent.)

### DIFF
--- a/dp_wizard/app/results_panel.py
+++ b/dp_wizard/app/results_panel.py
@@ -125,6 +125,7 @@ def make_download_or_modal_error(download_generator) -> str:  # pragma: no cover
             ui.pre(str(e)),
             title="Error generating code",
             size="xl",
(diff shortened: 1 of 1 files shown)


dp_wizard/app/results_panel.py
```diff
@@ -125,6 +125,7 @@ def make_download_or_modal_error(download_generator) -> str:  # pragma: no cover
             ui.pre(str(e)),
             title="Error generating code",
             size="xl",
+            easy_close=True,
```
Shoeboxam, line 128: Slick!

Shoeboxam (approved):
Usability x100!

PR #386, opened 2025-04-16
Pin all dependencies for application install

### LINKED ISSUE #370: `pip install dp_wizard[app]`
An extra (perhaps called `app`?) could supply pinned versions of all dependencies. This would let us keep the loose pins for library install, but also provide an install which is much less likely to ever break.

mccalluc:
- Fix #370
- Also warn if minimum python version isn't satisfied

I'm sure there is wheel reinvention here, but I think it does what we want.


ekraffmiller (approved):
looks good, approved

PR #387, opened 2025-04-17
always show epsilon slider endpoints

### LINKED ISSUE #385: Epsilon tag 10.0 disappears once the slider goes to >= 4

just like the image, `10.0` should be there but disappear somehow
Image

mccalluc:
- Fix #385

(In defense of the off-the-shelf widget, the end values are being hidden because there should be another box containing the current value over the slider, and when they would overlap, the endpoints are suppressed. At some point, it might be better to actually make a log scaled slider, instead of CSS hacking the linear scale.)


ChengShi-1 (approved):
check it on the page, it works well now

PR #388, opened 2025-04-17
Let user just provide column names

### LINKED ISSUE #368: Let user provide just the column names
so they don't need to upload CSV.

Do a design first and get feedback.

- Prereq for #362

### LINKED ISSUE #389: Remove underused CLI flags
```
--public_csv CSV
--private_csv CSV
--contrib CONTRIB
```
Initially I was thinking it would be nice to be able to set the entire initial state from the CLI, but now I'm feeling that that is clutter, particular if the number of options grows.

Having all these options also makes inconsistent states possible, and the error checking that is necessary to clean that up shouldn't be our priority.

mccalluc:
Discussed UI with Mike before starting. Considered having the UI more stable between the modes, and just disabling the sections which are not available, but decided that would be more confusing than doing it this way.

- Fix #368
- Fix #389


README-PYPI.md
```diff
@@ -25,27 +25,28 @@ You can check your current version with `python --version`.
 The exact upgrade process will depend on your environment and operating system.
 
 ```
-usage: dp-wizard [-h] [--public_csv CSV] [--private_csv CSV] [--contrib CONTRIB] [--demo]
+usage: dp-wizard [-h] [--demo] [--no_uploads]
```
mccalluc, line 28: These CLI options we're being used, and they created the possibility for inconsistent states. (What happens if `--private_csv` and `--no_uploads` are specified together?) Rather than adding complex validation logic, just get rid of features no one needs.

dp_wizard/app/__init__.py
```diff
@@ -31,13 +34,76 @@ def ctrl_c_reminder():  # pragma: no cover
     print("Session ended (Press CTRL+C to quit)")
 
 
+def _make_demo_csv(path: Path, contributions):
```
mccalluc, line 37: This function is just being moved down, closer to where it is needed, and the test is brought in as well. (This is a little big for a doctest, but if we need to tweak the columns, it will nice not to need to switch between files.)

dp_wizard/app/__init__.py
```diff
@@ -31,13 +34,76 @@ def ctrl_c_reminder():  # pragma: no cover
...
+    10
+    """
+    return max(min(n, upper_bound), lower_bound)
+
+
 def make_server_from_cli_info(cli_info: CLIInfo):
     def server(input: Inputs, output: Outputs, session: Session):  # pragma: no cover
-        public_csv_path = reactive.value(cli_info.public_csv_path or "")
-        private_csv_path = reactive.value(cli_info.private_csv_path or "")
+        if cli_info.is_demo:
```
mccalluc, line 92: This information used to be provided by the `CLIInfo`, but I think it works better to have these details specified here.

dp_wizard/app/results_panel.py
```diff
@@ -33,89 +33,7 @@ def button(name: str, ext: str, icon: str, primary=False):
     )
 
 
-def results_ui():
```
mccalluc, line 36: Most of the details remain the same, but since at the top-level it is now conditional, it has to be in the server.

tests/test_app.py
```diff
@@ -15,12 +15,20 @@
...
+default_app = create_app_fixture(Path(__file__).parent / "fixtures/apps/default_app.py")
 tooltip = "#private_csv_path-label svg"
 for_the_demo = "For the demo, we'll imagine"
 
 
+def test_cloud_app(page: Page, cloud_app: ShinyAppProc):  # pragma: no cover
+    page.goto(cloud_app.url)
+    expect(page).to_have_title("DP Wizard")
+    expect(page.get_by_text("Choose Public CSV")).not_to_be_visible()
+    expect(page.get_by_text("CSV Column Names")).to_be_visible()
```
mccalluc, line 29: Should add more here! But doesn't need to block review of everything else.

README-PYPI.md
mccalluc: Considering this a bit more, `demo` and `no_uploads` really should be mutually exclusive. I'll fix this.

ekraffmiller:
HI @mccalluc I just started to look at this and it works well, but it occurred to me that we aren't handling the possibility of string values in a column.  In DPCreator, we allowed users to create histograms by providing the string values of each category, rather than a lower and upper bound.  Also we had logic to prevent users from choosing numeric analysis types like mean and median for string data.  
I know it's probably for a separate ticket, but just wanted to bring it up here. Since we aren't handling strings, it would be good to at least have help text that explains that.

mccalluc:
>  In DPCreator, we allowed users to create histograms by providing the string values of each category, rather than a lower and upper bound. 

Filed:
- #393
- #394

I was thinking of just changing the prompt 
```diff
CSV Column Names
CSV Numeric Column Names
```
... but that's unnecessarily restrictive: We could use a string column for grouping, even if we can't take a mean.

I will change the text to indicate that we just need the columns to be used in the analysis, not all the columns.

ekraffmiller:
I'm not sure the best UI solution, but what I was trying to avoid is an unexpected error when the user chooses histogram for a column of strings, for example:

Screenshot 2025-04-22 at 10 35 16 AM

In the case of the "no_upload" mode, for histograms, the user could have the option of entering upper and lower bounds (for numbers only) or categories (which would work for numbers or strings).

mccalluc:
@ekraffmiller :

> the user could have the option of entering upper and lower bounds (for numbers only) or categories (which would work for numbers or strings).

I have filed 
- #393 

I hope discussion of that feature can be moved there; it will require several steps:
- UI design for eliciting column types.
- Implement that UI
- Use that new information to make the histogram UI conditional, both in its sidebar inputs, and in the preview.

This really feels like a separate issue: `--no_uploads` is strictly less powerful than what we have now, for safety's sake, while histograms on string values is a _new_ feature which depends on getting _more_ information from the user and introducing _more_ conditional UI.

Thinking about the overall project timeline, I would really like to use an on-line version of DP Wizard for my workshop June 3. I think we are agreed that for an on-line DP Wizard, we want to be sure that users aren't uploading files: This PR is necessary. I also have two weeks of vacation between now and then, so it will really help me if this PR can be considered on its own merits. Thank you.

ekraffmiller:
> @ekraffmiller :
> 
> > the user could have the option of entering upper and lower bounds (for numbers only) or categories (which would work for numbers or strings).
> 
> I have filed
> 
> * [Histograms of string-valued columns #393](https://github.com/opendp/dp-wizard/issues/393)
> 
> I hope discussion of that feature can be moved there; it will require several steps:
> 
> * UI design for eliciting column types.
> * Implement that UI
> * Use that new information to make the histogram UI conditional, both in its sidebar inputs, and in the preview.
> 
> This really feels like a separate issue: `--no_uploads` is strictly less powerful than what we have now, for safety's sake, while histograms on string values is a _new_ feature which depends on getting _more_ information from the user and introducing _more_ conditional UI.
> 
> Thinking about the overall project timeline, I would really like to use an on-line version of DP Wizard for my workshop June 3. I think we are agreed that for an on-line DP Wizard, we want to be sure that users aren't uploading files: This PR is necessary. I also have two weeks of vacation between now and then, so it will really help me if this PR can be considered on its own merits. Thank you.

@mccalluc yes, of course we can address this in the other issue.  I was just continuing the discussion here to avoid confusion, but maybe I created it instead.  I agree, it's a big change.

PR #392, opened 2025-04-21
Notebook execution works on windows

mccalluc:
Draft; Still need to handle single vs. double quoting in shell, at least.


mccalluc:
I was hoping for failures on windows CI: I think the best course might be to borrow a windows computer and reproduce the problem locally. Keep as draft.

PR #404, opened 2025-04-24
Run DP wizard in the cloud

### LINKED ISSUE #362: Run dp-wizard in the cloud?
For some users, one barrier might be installation. We could offer a lot of dp-wizard functionality without reading any private data:
- #368
- at the end, we can't execute the notebook, but we could still generate code they could then run locally.

mccalluc:
- Fix #362

It is currently on the free tier, which has 20 hours of compute per month. I would like this to be under Stefano's plan, but that doesn't need to block. 

I have asked how we can get a CNAME record for this, so we can have a shorter url, but that also does not need to block.


mccalluc:
> I tried then cloud version, and I got an error trying to download the PDF:

Good catch! I've tripped over this before -- I think this is a dependency that is needed at run time, but is provided by the test library itself, so CI doesn't catch it. In my court.

mccalluc:
A little bit farther, and now I get a different error:
> No suitable chromium executable found on the system. Please use '--allow-chromium-download' to allow downloading one,or install it using `playwright install chromium`

As I understand it, PDF rendering requires a whole headless browser that loads the html and prints it to PDF.

This is worth resolving: I think these are issues that could affect the local install, too... but if it goes on too long, we could also consider dropping PDF export. User could always print the HTML to a PDF.

mccalluc:
I'm going to go in a different direction here: Even if I can get the chromium download sorted out, that will be a blocking operation. It looks like the platform builds and caches an image with everything in `requirements.txt`, but I don't know of a way to get installs cached that aren't part of `requirements.txt`.

mccalluc:
At this point, I'd like to drop the PDF rendering altogether: It felt like it would be an easy win, but it's turned out to be something that continually requires developer time.

dp_wizard/shiny/dataset_panel.py
```diff
@@ -9,7 +9,7 @@
     PUBLIC_PRIVATE_TEXT,
 )
 from dp_wizard.utils.csv_helper import get_csv_names_mismatch
-from dp_wizard.app.components.outputs import (
+from dp_wizard.shiny.components.outputs import (
```
mccalluc, line 12: Add more welcome info on the dataset panel:
- "DP Wizard" title
- Relationship to the OpenDP project
- Link to github / pypi?

ekraffmiller: > Add more welcome info on the dataset panel:
> 
> * "DP Wizard" title
> * Relationship to the OpenDP project
> * Link to github / pypi?

I don't see any changes in the dataset panel, are you planning on doing that in this PR?

ekraffmiller:
> At this point, I'd like to drop the PDF rendering altogether: It felt like it would be an easy win, but it's turned out to be something that continually requires developer time.

I think that's fine, the other formats should be sufficient

mccalluc:
Latest changes:
- Drop PDF rendering everywhere
- When cloud deployed, more of a welcome message
- Add deployment script, and move all scripts to sub-directory: Top level is getting cluttered.
- Rename from `no_uploads` to `in_cloud`, since it is more than just disallowing uploads.

ekraffmiller (approved):
looks good, approved!

PR #412, opened 2025-05-04
Upgrade h11

mccalluc:
I pinned it in `requirement.in`, rebuilt, removed the pin, and ran again, so this is a consistent state.

Another approach would be to leave the pin in `requirements.in`: I could do that, if you'd prefer.

- Fix https://github.com/opendp/dp-wizard/security/dependabot/10
- Fix https://github.com/opendp/dp-wizard/security/dependabot/8


ChengShi-1 (approved):
looks good to me

PR #414, opened 2025-05-05
Use function templates for short blocks

### LINKED ISSUE #309: Apply function-template machinery
Follow-up to:
- #301

Try to move all the stand-alone templates into the main files.

mccalluc:
- Fix #309

The remaining template files are larger, so it makes more sense to leave them where they are.

Motivation: Easier to maintain with these inline, instead of linking out to small files.... I think?

For the reviewer:
- Is having these short blocks inline actually an improvement? It does save the trouble of switching between files, but it is not a common pattern.


ChengShi-1:
Hi @mccalluc , would you mind taking a look at the test failure?

mccalluc:
Failure is in an end-to-end test: these can be flaky. Re-running. See if it works...

ChengShi-1 (approved):
I think it is more straightforward now, so looks good to include these changes

PR #415, opened 2025-05-05
test error handling in notebook execution

### LINKED ISSUE #333: Test stack trace in UI if notebook fails
If there is a problem with the notebook, after #312 we show an error in the UI. To test this we either need to 
- add a back door that lets us generate an intentionally bad notebook (bad)
- make a mini-application just to test this case (more work)

Not a top priority, but still a good thing to do.

> I don't think its necessary now to add a test for it, but we could make it an issue to work on later.

_Originally posted by @ekraffmiller in https://github.com/opendp/dp-wizard/pull/312#pullrequestreview-2736271726_

mccalluc:
- Fix #333

Adds an undocumented CLI option that overrides the content of the notebook with a simple `raise Exception`. If we need other backdoors for testing, we might change this from a boolean to a string, but that's not needed yet.


ChengShi-1 (approved):
Looks good and works well.

PR #416, opened 2025-05-05
informative download file names

### LINKED ISSUE #313: Title download to reflect analysis choices
If it's run multiple times, the user has multiple downloads, all with the same name.

There's a lot of information which could be included: What is most important?
- Input CSV name?
- Date/time?
- Columns selected?
- Analysis choices?

Or we could let the user pick:
- on the downloads tab, add a free text field with a prepopulated value?
- or a set of checkboxs for the fields to use?

Besides the filename, the title in the notebook could also be updated, and/or add a markdown table with a summary of the input choices?

mccalluc:
- Fix #313
- Doesn't include grouping, or any of the parameters, but I don't think it should include everything. feedback welcome!
- one mime type fix

### DIFF
--- a/dp_wizard/shiny/results_panel.py
+++ b/dp_wizard/shiny/results_panel.py
@@ -204,7 +204,7 @@ def results_server(
         )
 
     @reactive.calc
-    def analysis_download_stem() -> str:
+    def download_stem() -> str:
         description = ", ".join(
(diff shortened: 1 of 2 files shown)


ekraffmiller (approved):
Good idea, it's helpful to have different file names for different runs of the app.

PR #417, opened 2025-05-05
Show warnings if you jump around tabs

### LINKED ISSUE #32: After release, controls on previous pages should be locked

mccalluc:
- Fix #32 

... though not actually locking. I think we're ok with a softer model.
- For the tabs where we haven't provided the prereqs, it's still useful to be able to look around and see what the rest of the app looks like.
- For after the release, we want to point out that the epsilon is spent, but since we can't keep people from just restarting the application, it doesn't make sense to lock it down hard.

### DIFF
--- a/dp_wizard/shiny/__init__.py
+++ b/dp_wizard/shiny/__init__.py
@@ -111,6 +111,7 @@ def make_server_from_cli_info(cli_info: CLIInfo):
         groups = reactive.value([])
         weights = reactive.value({})
         epsilon = reactive.value(1.0)
+        released = reactive.value(False)
 
         about_panel.about_server(
             input,
@@ -121,6 +122,7 @@ def make_server_from_cli_info(cli_info: CLIInfo):
             input,
             output,
             session,
+            released=released,
             is_demo=cli_info.is_demo,
             in_cloud=cli_info.in_cloud,
             initial_public_csv_path="",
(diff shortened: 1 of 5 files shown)


ekraffmiller:
Rather than the warnings, can we prevent the user from performing an action out of order?  For example, in the Download Results tab, if I click Download notebook before doing an analysis, I get an error.  Instead of showing the warning, can we disable the button? 

Screenshot 2025-05-06 at 1 44 01 PM

mccalluc:
@ekraffmiller : That's a good catch. Just filed:
- #420

I think the messaging proposed in this PR would also be good:
- Warning about (but not blocking?) another release after the first.
- Explaining why buttons on a tab are disabled.

If you're not sure, might be good a good design question to sort out at meeting.

ekraffmiller:
> @ekraffmiller : That's a good catch. Just filed:
> 
> * [Disable downloads until requirements are satisfied #420](https://github.com/opendp/dp-wizard/issues/420)
> 
> I think the messaging proposed in this PR would also be good:
> 
> * Warning about (but not blocking?) another release after the first.
> * Explaining why buttons on a tab are disabled.
> 
> If you're not sure, might be good a good design question to sort out at meeting.

Sure I think it's fine to keep the warnings, along with disabling where appropriate 👍

ekraffmiller (approved):

PR #421, opened 2025-05-06
Specify path instead of actually uploading file

### LINKED ISSUE #314: Alternative to file upload widget
For a local application, the file upload idiom is less than ideal:
- The potentially sensitive file is copied to a new location
- Is this temp copy deleted when the application exits?
- The generated notebook has a hard coded reference to the temp file

Could we make an alternative widget that supports navigation of the local filesystem and returns a path, without copying the file?

mccalluc:
- Fix #314

WIP!


mccalluc:
Asked for help on the forum: https://forum.posit.co/t/alternative-to-ui-input-file-when-i-just-want-the-path-not-the-contents-of-the-file/200758

mccalluc:
Status update: I could imagine users not liking the current file upload behavior, so I'd like to keep this around just to remind us of the alternative.

PR #422, opened 2025-05-06
Disable downloads if analysis undefined

### LINKED ISSUE #420: Disable downloads until requirements are satisfied
- From #417

>  Rather than the warnings, can we prevent the user from performing an action out of order?  For example, in the Download Results tab, if I click Download notebook before doing an analysis, I get an error.  Instead of showing the warning, can we disable the button? 

Screenshot 2025-05-06 at 1 44 01 PM

_Originally posted by @ekraffmiller in https://github.com/opendp/dp-wizard/issues/417#issuecomment-2855439227_

Let me start a new PR just for this.

mccalluc:
- Fix #420

This PR doesn't add any messaging about _why_ the buttons are disabled: That is handled in #417.

### DIFF
--- a/dp_wizard/shiny/results_panel.py
+++ b/dp_wizard/shiny/results_panel.py
@@ -20,16 +20,25 @@ from dp_wizard.utils.converters import (
 wait_message = "Please wait."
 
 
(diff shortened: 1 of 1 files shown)


ekraffmiller (approved):
looks good, approved

PR #430, opened 2025-05-08
use templated expr

mccalluc:
- Towards #219 

The root problem is that details are duplicated between the templates and `db_helper.py`. The original thought was that db_helper code could be moved into shared, but that by itself wouldn't fix the problem: the templates are still necessary. Putting everything in shared would result in less readable, less idiomatic generated code.

The course proposed here is to `exec` the templated code, and in some sense there is less duplication now... but just looking at the diff, it's hard to argue that this is _actually_ simpler.

For the reviewer: Is it better to stick with the status quo, and tolerate some duplication of code, and close this PR without merging, and close the linked issue... or add complexity with `exec`, but have the polars expression in only one place? I'm on the fence.

### DIFF
--- a/dp_wizard/shiny/results_panel.py
+++ b/dp_wizard/shiny/results_panel.py
@@ -47,7 +47,7 @@ def button(
 
 def _strip_ansi(e):
     """
-    >>> e = Exception('\x1B[0;31mValueError\x1B[0m: ...')
+    >>> e = Exception('\x1b[0;31mValueError\x1b[0m: ...')
     >>> _strip_ansi(e)
     'ValueError: ...'
     """
--- a/dp_wizard/utils/dp_helper.py
+++ b/dp_wizard/utils/dp_helper.py
@@ -1,12 +1,39 @@
 import polars as pl
 import opendp.prelude as dp
 
-from dp_wizard.utils.shared import make_cut_points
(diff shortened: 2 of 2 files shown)


mccalluc:
CI failure. Re-running:
```
Run actions/upload-artifact@v4
  with:
    name: playwright-traces
    path: test-results/
    if-no-files-found: warn
    compression-level: 6
    overwrite: false
    include-hidden-files: false
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.13.3/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.13.3/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.13.3/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.13.3/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.13.3/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.13.3/x64/lib
With the provided path, there will be 1 file uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

PR #431, opened 2025-05-08
In means plot title, name the grouping

### LINKED ISSUE #344: Grouped DP means plot should label axis and/or describe grouping
It doesn't describe what we're grouping by in the plot.
```
if groups:
    title = (
        f"DP means for grade, "
        f"assuming {contributions} contributions per individual"
    )
    plot_bars(grade_stats, error=0, cutoff=0, title=title)
```
Axis label might be better than title, though that might require more changes to `plot_bars`

mccalluc:
- Fix #344


mccalluc:
Let me do some more manual testing on this. `groups` might be used in the histogram in a way that conflicts.

PR #437, opened 2025-05-08
more demo defaults

### LINKED ISSUE #436: Set more demo defaults
We should be be able to just fill in the whole analysis plan:
- histogram
- upper bound to 100

mccalluc:
- fix #436
Not working yet. Still draft.


mccalluc:
I put a breakpoint on `cleanup_reactive_dict`, and it's being called on the initial load, from `_on_columns_change`. The dict does have a good initial value, but then it is cleared... I think because it's trying to be synchronized with the UI, and the UI starts empty.

PR #440, opened 2025-05-09
generalize analysis plans: dict of lists

### LINKED ISSUE #434: Generalize AnalysisPlan data structure to allow multiple analyses for a given column
At some point we'll want to support multiple analyses per column. (I'm thinking particularly about DP Regression.) Before any UI work can happen, we'll need to generalize the current data structure: Right now it's a dictionary, and I think the first step would be just to make the values lists of analyses, rather than just an analysis at the top level.

mccalluc:
- Fix #434

For the reviewer:
- Conceptually, does this seem like a good data structure? Alternatives might be more structure in the key (perhaps a tuple?) or a dict of dicts, instead of a dict of lists... but those both seem more complicated.

### DIFF
--- a/dp_wizard/shiny/results_panel.py
+++ b/dp_wizard/shiny/results_panel.py
@@ -47,7 +47,7 @@ def button(
 
 def _strip_ansi(e):
     """
-    >>> e = Exception('\x1B[0;31mValueError\x1B[0m: ...')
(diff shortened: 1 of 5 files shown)


ChengShi-1 (approved):
I agree with this structure, because a dict of lists is natural and intuitive

PR #441, opened 2025-05-09
Note about "None" for string valued columns

### LINKED ISSUE #394: Usability with string-valued columns
See also
- #368

On #388 Ellen writes:
>  Also we had logic to prevent users from choosing numeric analysis types like mean and median for string data.

I think this is blocked on work upstream in OpenDP:
- https://github.com/opendp/opendp/issues/2374

Until that is addressed, there's nothing for us to do if we did collect types from the user... and once it is addressed, we will be required to collect types.

mccalluc:
- Fix #394 
- Update the demo CSV and tooltips: use a string column for grouping.

Does this seem sufficient to close the issue? At notebook generation time we can't look at the data, so we can't tell whether there are `None`s or not.

A couple times the idea of prompting for the types of columns has been suggested, and we might end up there, but I'd like to understand the usecases before we jump into implementing that. I think there are at least two cases to be distinguished:

- User knows their data and marks a column as string. 👉 We don't let them do a numeric analysis on this column.
  - ... but for me, this doesn't seem like a good motivation, because if they understand the column is a string, I'd trust them not to try to take the mean: not worth adding UI complexity.
- or: user doesn't know that there are strings lurking in the column 👉 so we can't trust the types supplied by the user
  - ... and we still need to explain odd downstream results.

Maybe there are other use cases you have in mind, or maybe your analysis is different?

for the reviewer:
- If this is better than the status quo, please approve.
- If you think this resolves the linked issue, great!
- If this is ok, but doesn't really resolve the issue, that's good too: I can unlink the issue, and we can discuss what a real fix would look like.... but hopefully not in this PR.

### DIFF
--- a/dp_wizard/shiny/__init__.py
+++ b/dp_wizard/shiny/__init__.py
@@ -47,27 +47,34 @@ def _make_demo_csv(path: Path, contributions):
     ...         rows = list(reader)
     ...         rows[0].values()
     ...         rows[-1].values()
-    ['student_id', 'class_year', 'hw_number', 'grade', 'self_assessment']
-    dict_values(['1', '2', '1', '82', '0'])
-    dict_values(['100', '2', '10', '78', '0'])
+    ['student_id', 'class_year_str', 'hw_number', 'grade', 'self_assessment']
+    dict_values(['1', 'sophomore', '1', '82', '0'])
+    dict_values(['100', 'sophomore', '10', '78', '0'])
     """
     random.seed(0)  # So the mock data will be stable across runs.
     with path.open("w", newline="") as demo_handle:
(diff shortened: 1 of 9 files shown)


ekraffmiller:
I think adding the note is an improvement, but I think there is still enhancements that could be made related to this issue ([usability with string valued columns](https://github.com/opendp/dp-wizard/issues/394)). For example, for histograms, it would be helpful for the user to have the option of providing categories for bins, if the data type is string.  To keep it simple, we wouldn't have to require that the user enter the data types of all columns, we could just add something to the histogram analysis that would allow the user to either enter categories, or max/min bin values.
But since that is a separate issue I will approve this. 👍

ekraffmiller (approved):
looks good, approved

PR #451, opened 2025-05-12
Fill in templates for medians

### LINKED ISSUE #275: Fill in the visualization/report/etc. for means and medians

mccalluc:
- Fix #275 ... actually, means was already done.
- Prompt for number of candidates

Motivation: The generated code for medians only included a `TODO` in some places, instead of actually completing the analysis. This is debt.


ekraffmiller:
This works well, the only thing I noticed is that there doesn't seem to be validation on the median analysis, I was able to enter these values: 
Screenshot 2025-05-28 at 4 36 54 PM

ekraffmiller (approved):
looks good, approved

PR #454, opened 2025-05-12
Suppport count queries

### LINKED ISSUE #393: Histograms of string-valued columns
On #388, Ellen writes:

>  In DPCreator, we allowed users to create histograms by providing the string values of each category, rather than a lower and upper bound. 

I think the best way to support this would be to add plain counts as an analysis option. The user could then group by a string column, and then pick anything else and count it... but count does feel a little different because it really doesn't depend on the value in the column.

When someone wants a histogram, it might be hard not to select the "histogram" option. There's a usability question here.

mccalluc:
- Fix #393

Motivation: Count is a useful, basic statistic in DP, even if we tend to take it for granted.

Count is a little different, because we don't actually do anything with the value. We could count non-null values, but my sense is that adds to the complexity without making it more useful.


mccalluc:
The estimated rows isn't used to create the notebook, and there's some sense that this is confusing: Users reasonably want to know what goes into the notebook's calculation, and what doesn't. Onyinye is going to gather more feedback on that.

Test failure might be a flaky end-to-end test. Rerunning now.

mccalluc:
Looked at the playwright artifact, and the failure is the same one other tests have hit:
- #510

I should fix it! But for now, just rerunning tests.

PR #460, opened 2025-05-28
plan_column now a list

### LINKED ISSUE #459: Notebook download: `'list' object has no attribute 'analysis_type'`
This is left over from #440. (Maybe an intersection with another PR? not quite sure.) It has been fixed in other PRs, just to make CI happy, but in case those are blocked, this is a small fix for just one bug.

mccalluc:
- Fix #459

Other PRs incorporate this fix as part of conflict resolution, but if they are stalled, this could be merged.


mccalluc:
Handled by 
- #451

PR #462, opened 2025-05-30
deployment fixes

### LINKED ISSUE #461: Make corrections to posit deploy instructions

mccalluc:
Formerly, it was just pushing the current state of `cloud-deployment` without picking up any changes on main.

- fix #461

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -24,5 +24,3 @@ echo "Push..."
 git push -f origin main:cloud-deployment
 
 echo "Redeployed!"
-
-git checkout main


Shoeboxam (approved):
The changes seem reasonable. As I understand it, you want to ensure that deploys only happen from main? But doesn't that check already happen on line 16?

mccalluc:
> The changes seem reasonable. As I understand it, you want to ensure that deploys only happen from main. But doesn't that check already happen on line 16?

I guess I also want to make sure that the test run is on main.

Without the "main:", it was just pushing `cloud-deployment`, but without committing any of the changes from `main` onto that branch.

(I could skip this script entirely and just set `main` as the branch that is monitored for updates, but that seems a little fragile.)

PR #449, opened 2025-05-10
one dependency per line in toml

mccalluc:

PR #478, opened 2025-06-10
Update dependencies with dependabot alerts

mccalluc:
Suggest reviewing and merging this first:
- #449

It will make the toml diff less ugly


ekraffmiller (approved):
looks good, approved

ekraffmiller (approved):

PR #488, opened 2025-06-17
Add an easy-to-read HTML report

### LINKED ISSUE #485: Easy-to-read HTML report
We want a report style that emphasizes readability. Currently the only HTML output is the notebook.

- graphs
- tables
- confidence intervals

mccalluc:
- Fix #485 

WIP. Needs a spec for the HTML report.

### DIFF
--- a/dp_wizard/utils/code_generators/no-tests/_reports.py
+++ b/dp_wizard/utils/code_generators/no-tests/_reports.py
@@ -34,8 +34,21 @@ def save_html_report(report):
     from htmltools import tags
     from yaml import dump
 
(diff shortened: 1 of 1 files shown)


mccalluc:
Not sure about the best way to add plotting here.
- just put the data in the HTML, and use D3 or similar to render the chart. Downside is that this chart will look different from the preview.
- Or in `shared.py`, add a call to https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.savefig.html, and have some kind of global state that is accumulating plots, and then we incorporate them (via data-URIs) into the final html document?

PR #489, opened 2025-06-18
validate number of bins; generate dummy CSV when running in cloud mode

### LINKED ISSUE #468: "Number of Bins" validation
- Negative values can be entered without error
- "0" produces `float division by zero`
- Positive floats produce no error... but maybe this is ok
- "asdf" produces `int() argument must be a string, a bytes-like object or a real number, not 'NoneType'`

### LINKED ISSUE #473: Notebook could generate a csv if one wasn't provided.
Something like:
```
# TODO: Delete this and change the path to your CSV.
from pathlib import Path
Path('fill-in-correct-path.csv').write_text('grade\n100')
```
Take the midpoint of the bounds?

mccalluc:
- fix #468

The error message isn't what I'd really like, but we're limited by the information available, unless we want to add a backchannel. I don't think the additional complexity is worth it.

- fix #473

Pushed to the same PR by accident. If you care strongly, I can split this to a separate PR.


ChengShi-1:
image
Maybe we could also add a constraint for error message when there is a too large number?

mccalluc:
@ChengShi-1 , this addresses the bug you reported, and also handling small ranges better. (Like 0 to 0.001.) I'm validating the input in the UI, and also making the code in `shared.py` more robust... at the cost of a little more complexity in the generated notebook. I think it's a good balance, but feedback welcome.

Issue #498, opened 2025-06-26
Draft a UI to help users understand unit of privacy


mccalluc:
- Decide on the kind of thing that is being protected
  - For example, individual-day / individual / household
  - (This isn't actually used downstream)
- and then, the number of rows they can touch

Given those two, we could make an info graphic.

mccalluc:
What we have now:
> How many rows of the CSV can one individual contribute to? This is the "unit of privacy" which will be protected.
> 
> Contributions
> [______________]

Proposal:

First prompt for the entity to protect. This isn't actually used in any calculations downstream, but it prompts the user to think more concretely about what they are protecting.

> To apply differential privacy we need to identify the entity whose privacy we are protecting.
>
> (_Select list with longer description shown for the current selection; Default to "Individual"._)
>
>  📅 Individual day: "DP can also be used to protect an individual's data for some span of time. This can give more accurate results, and may also be more tractable if you don't have a unique ID for individuals."
> 👤 Individual: "DP often is used to protect the privacy of individuals, but but depending on your needs you might want a less granular or more granular entity."
> 🏠 Household: "If the privacy of a member of my family is violated, I may consider my own privacy to be violated. We might want to protect a less granular entity than the individual."

Then, once we have an entity, we can ask how that entity is represented in the CSV:
> How are individuals (_or individual days, or households_) represented in your CSV?
> How many rows can one indiviual contribute?
> [______________]
>
> Alternatively, which columns taken together comprise a unique key for individuals?
> [______________]

PR #502, opened 2025-06-27
Analysis inputs empty by default

### LINKED ISSUE #501: Set inputs to empty on load
Users should be intentional about filling in values.

mccalluc:
- Fix #501

The root problem is that we've seen users glide through without taking time to change bounds. If bounds start off empty, and there's a validation error message, then they need to make a conscious choice before moving forward.

Showing the validation error message on load might be a little weird, but if it reads as a todo list rather than a you-did-something-wrong message, it probably works? **Would appreciate feedback on this!**

This is only checking the bounds right now. #489 adds validation for bin count, but the framework doesn't let numeric values be empty, and setting it to zero or one might just be more confusing: I hope people will see the 10 bars in the bar chart, and they can make their own decision if that's right for them. **Would it be good to change bin count to a string so it could also start off empty?** Would lose the increment/decrement buttons in the UI.

I have a separate issue (#498) to work on the unit of privacy UI.

### DIFF
--- a/dp_wizard/shiny/__init__.py
+++ b/dp_wizard/shiny/__init__.py
@@ -112,6 +112,7 @@ def make_server_from_cli_info(cli_info: CLIInfo):
 
         public_csv_path = reactive.value("")
         analysis_types = reactive.value({})
+        analysis_errors = reactive.value({})
         lower_bounds = reactive.value({})
         upper_bounds = reactive.value({})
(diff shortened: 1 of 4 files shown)


mccalluc:
Add "numeric"!

ekraffmiller (approved):
validation for the bounds input looks good, approving

PR #506, opened 2025-06-30
Make the epsilon slider finer grained

### LINKED ISSUE #497: Reduce step-size in epsilon slider
Onyinye's users really wanted to get to particular users. Providing an alternate input mode would be a big lift, but reducing the step size is easy.

mccalluc:
- Fix #497

With this change, increments are not quite down to the pixel level, but they are small.

The user can get much closer to some integer values, which I think is one of the main desires.

- 2.0
- 3.02
- 3.98
- 5.01
- 6.03

I feel some ambivalence about this. If we need to give users more precise control, it really should be over the error margin, not the epsilon.

### DIFF
--- a/dp_wizard/shiny/components/inputs.py
+++ b/dp_wizard/shiny/components/inputs.py
@@ -41,5 +41,5 @@ def log_slider(id: str, lower_bound: float, upper_bound: float):
 </style>
 """
         ),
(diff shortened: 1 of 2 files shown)


mccalluc:
Went a different way:
- #575

Closing this PR because we actually want to discourage fine grained control for a few reasons:
- Doing it with a slider is just frustrating
- Every bump is a recalculation, and we want to show the bigger changes
- Longer range, we should be thinking about accuracy, not epsilon

PR #509, opened 2025-06-30
Add note about grouping in panel titles

### LINKED ISSUE #496: Include grouping info in the titles of the analysis panes
Each analysis is really doing something a little different when grouping is applied, so it's good to be reminded of that.

mccalluc:
Screenshot 2025-07-16 at 10 30 06 AM

- Fix #496
- Also add a note below, explaining that their will be a bar chart when the data is grouped.

For the reviewer:
- Does this text seem like a good way to explain what's going on?
- Is there a way we could do this better, _without_ text?


ekraffmiller (approved):
looks good, approved

PR #512, opened 2025-07-01
inline playwright helper functions

### LINKED ISSUE #511: Playwright helper functions less than helpful
These functions
```
    def expect_visible(text):
        expect(page.get_by_text(text)).to_be_visible()

    def expect_not_visible(text):
        expect(page.get_by_text(text)).not_to_be_visible()

    def expect_no_error():
        expect(page.locator(".shiny-output-error")).not_to_be_attached()
```
actually just get in the way when trying to debug: If I'm stepping through the code, it's jumping between frames, and it's harder to see what the assertions are.

Just inline these. We're not using helper functions on the other app tests, and they are fine.

mccalluc:
- Fix #511


ChengShi-1 (approved):
Looks good.

PR #516, opened 2025-07-01
Pin doc links

### LINKED ISSUE #515: Pin docs links to the OpenDP version used
When users navigate from the notebook to the docs, we should make sure they are looking at a version of the docs that matches the version of the library that the generated notebook uses.

mccalluc:
- Fix #515


ChengShi-1 (approved):
Good! 
I just check the notebook and the changes work well.

PR #522, opened 2025-07-01
Add note about "dp.median" in generated code

### LINKED ISSUE #500: Add a note that quantile in notebook is being used for median
Someone was looking for "median" and was confused not to see it.

mccalluc:
- Fix #500


ekraffmiller (approved):
looks good, approved

PR #531, opened 2025-07-08
Add categorical histogram

mccalluc:


mccalluc:
- The linked issue doesn't really describe what this does
- Since this PR, we've added more type information, so analyses are only selectable for numeric columns
- If this is needed, probably better to spin this off from counts

PR #540, opened 2025-07-18
estimated rows should not clear bounds

### LINKED ISSUE #537: Changing estimated rows clears bounds
I think this is a regression. Probably just need to make it reactive.

mccalluc:
- fix #537

Draft for now: I'm seeing a few problems with UI elements being reset, and I'll try to fix them all in this PR, and then mark ready for review.


mccalluc:
Replaced by #541

PR #542, opened 2025-07-31
Even more typing

mccalluc:
- Extends #541 by adding more types.
- Instead of a bare wrapper around `str`, move the clean-up utils into the constructor, so we can be confident that any instance of these classes actually has the prescribed structure.


ekraffmiller (approved):
Looks good, the additional types make the code more readable 👍

PR #551, opened 2025-07-31
add backticks in notebook title; Move utility function into class

### LINKED ISSUE #549: Back-ticks around column name in title
- It's the style we use later in the notebook.
- It seems weird for the title to start with a lower case letter.

mccalluc:
- Fix #549


ekraffmiller (approved):
looks good, approved

PR #553, opened 2025-07-31
Add minimal test of last shared function

### LINKED ISSUE #546: Test `plot_bars`
```
def plot_bars(
    df: DataFrame, error: float, cutoff: float, title: str
):  # pragma: no cover
```
This one pragma note shows up in the generated notebook, so it'd be good to clean it up, and then add a check to make sure it doesn't come back.

mccalluc:
- Fix #546


mccalluc:
Redo from scratch.

PR #558, opened 2025-08-01
add a placholder, so both inputs are the same width

### LINKED ISSUE #556: UserWarning: More column widths
Fix this warning:
```
site-packages/shiny/ui/_layout_columns.py:209: UserWarning: More column widths than children at breakpoint 'sm', extra widths will be ignored.
```

(can we make warnings fatal during tests and dev?)

mccalluc:
- Fix #556


ekraffmiller (approved):
looks good, not getting the warning now.

PR #566, opened 2025-08-05
Consolidate state into dataclass

### LINKED ISSUE #563: Reactive for application state
The parameter lists get longer and longer, and when something needs to be added, it has to be added in several places. Make one object that contains all the details.

mccalluc:
Also organize the assignments to correspond to the structure of the application.

When a new property was added previously, all the signatures needed to be updated. There is still the work now of assigning to variables, but that seems lighter, and eventually we could drop the local variables and reference `state` everywhere, though I'm not sure right now that that would make things simpler.

- Fix #563


ekraffmiller (commented):
I think it helps the organization of the code to have the AppState class, looks good.

ekraffmiller (approved):

PR #569, opened 2025-08-06
Use dp-wizard-templates

### LINKED ISSUE #550: Pull out templating engine into separate library: `dp_wizard_templates`?
- The templating engine is mature: Haven't needed to touch it recently.
- It's kind of complicated.
- It might be useful for other projects.
- Publishing reusable code is one way to get visibility for the project.

mccalluc:
- Fix #550
- API changes slightly: Now need to supply a root directory for templates, instead of assuming a path from the current `__file__`
- Imports are messy: They will be tidied up, and have merge conflicts, with #565

Locally, code generation was failing for me, or rather notebook execution: It said it couldn't import polars? Error went away with a fresh install.

### DIFF
--- a/dp_wizard/utils/code_generators/abstract_generator.py
+++ b/dp_wizard/utils/code_generators/abstract_generator.py
@@ -1,8 +1,12 @@
-from dp_wizard import opendp_version, get_template_root
+from abc import ABC, abstractmethod
 from math import gcd
+from pathlib import Path
+from typing import Iterable
 
 from dp_wizard_templates.code_template import Template
 
+from dp_wizard import get_template_root, opendp_version
+from dp_wizard.types import ColumnIdentifier
 from dp_wizard.utils.code_generators import (
     AnalysisPlan,
     make_column_config_block,
@@ -11,12 +15,6 @@ from dp_wizard.utils.code_generators import (
 )
 from dp_wizard.utils.code_generators.analyses import histogram
 from dp_wizard.utils.dp_helper import confidence
-from dp_wizard.types import ColumnIdentifier
-
-from abc import ABC, abstractmethod
-from pathlib import Path
-from typing import Iterable
-
 
 root = get_template_root(__file__)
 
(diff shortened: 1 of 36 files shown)


dp_wizard/utils/code_generators/abstract_generator.py
```diff
@@ -70,8 +70,7 @@ def template():
             )
             .finish()
         )
-        # Line length determined by PDF rendering.
-        return black.format_str(code, mode=black.Mode(line_length=74))  # type: ignore
```
ekraffmiller, line 74: Why is the formatting removed here?

dp_wizard/utils/code_generators/analyses/count/__init__.py
```diff
@@ -11,9 +14,12 @@
 has_bins = False
 
 
+root = Path(__file__).parent / "no-tests"
```
ekraffmiller, line 17: `root `is defined in multiple places - could it be defined once and shared in the code?

ekraffmiller (commented):
Looks good! Just a couple of small comments.

dp_wizard/utils/code_generators/abstract_generator.py
mccalluc: I moved black formatting into DP Wizard Templates, but it is only applied to the final notebook render by default, and is optional when just doing a single template... but since we show code snips to users, it would be good to make sure it's still called here, through the library.

dp_wizard/utils/code_generators/analyses/count/__init__.py
mccalluc: Maybe. I'll need to check, but these are resolving to separate directories. (There are multiple `no-tests` directories.) Maybe this whole convention is confusing? Let me take another look...

mccalluc: Made a utility function, so hopefully it's a little less ugly?

dp_wizard/utils/code_generators/abstract_generator.py
mccalluc: actually, we only see this output once it has been transformed into a notebook, so black is applied automatically.

ekraffmiller (approved):
changes look good, approved

PR #503, opened 2025-06-27
Make the demo tooltips always-on

mccalluc:
- Fix #149
Screenshot 2025-06-27 at 4 34 15 PM

PR #571, opened 2025-08-08
tutorial toggle and sample.csv

### LINKED ISSUE #570: Demo-mode toggle
With the help-text more prominent:
- #503 
 
it seems a shame not to make use of it. Propose adding a toggle up top, next to dark mode.

mccalluc:
- Fix #570

Structurally, one change to note is that there's now a distinction between `is_demo_csv` and `is_demo_mode`. 
- `is_demo_csv` is a plain, non-reactive, bool and is set on the command line.
- `is_demo_mode` is reactive and can be turned on or off by the user. It starts on if we're running with `--demo` or `--cloud`.

Trying to avoid future merge conflicts, this PR includes changes from both of the following, so wait for them, and then merge main back to this branch:
- #566
- #503

### DIFF
--- a/dp_wizard/shiny/results_panel.py
+++ b/dp_wizard/shiny/results_panel.py
@@ -143,6 +143,14 @@ def results_server(
         disabled = not weights()
         return [
             ui.h3("Download Results"),
+            tutorial_box(
+                is_tutorial_mode(),
+                """
+                Now you can download a notebook for your analysis.
+                The Jupyter notebook could be used locally or on Colab,
+                but the HTML version can be viewed in the brower.
+                """,
+            ),
             ui.p("You can now make a differentially private release of your data."),
             # Find more icons on Font Awesome: https://fontawesome.com/search?ic=free
             ui.accordion(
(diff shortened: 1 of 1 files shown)


ekraffmiller (commented):
I think this looks great, except it's a little confusing to see references to the "quiz grades example" when I uploaded my own csv.  Can there be more logic in the tooltips, to only show text related to the demo file when it is being used?

Screenshot 2025-08-13 at 1 09 18 PM

mccalluc:
Good feedback. It's a little muddle right now, because there are two different "demos". Let me try to clear this up...

mccalluc:
@ekraffmiller : Please don't feel obliged to wade through the diff again! There is a lot of renaming, but I've cleaned up the outstanding references to "demo" and "tooltip"

I've made one message conditional on `sample.csv`... Could do it in more places, but I feel like a concrete example, even if it's not your domain, could be helpful in some cases?

ekraffmiller (approved):
looks good, approved!

PR #575, opened 2025-08-20
Round the epsilon values

### LINKED ISSUE #497: Reduce step-size in epsilon slider

mccalluc:
- Fix #497

Having the precise values of the exponential was a little frustrating to users, because it wasn't clear where these values were coming from. I hope this will be better, but the long-run plan is to focus more on accuracy than on epsilon, so I wouldn't overdo this.


ChengShi-1 (approved):
looks good

PR #580, opened 2025-08-22
change dataset page to two columns

mccalluc:
- Fix 579

Feedback on design welcome.
- The page was getting long, and particularly in wide windows, the space could be used more efficiently.
- Seems like the questions about the dataset could be brought together.

### DIFF
--- a/dp_wizard/shiny/dataset_panel.py
+++ b/dp_wizard/shiny/dataset_panel.py
@@ -69,19 +69,24 @@ def dataset_ui():
         "Select Dataset",
         ui.output_ui("dataset_release_warning_ui"),
         ui.output_ui("welcome_ui"),
-        ui.output_ui("csv_or_columns_ui"),
-        ui.card(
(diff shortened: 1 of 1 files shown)


ChengShi-1 (commented):
I like the two columns layout when I turn off the tutorial mode!

One suggestion, would it be better to extend the width of the `Welcome to DP Wizard, from OpenDP.` and other tooltips on the top of pages, so that it matches the width of the two columns?
image
like this
image

mccalluc:
> One suggestion, would it be better to extend the width of the Welcome to DP Wizard, from OpenDP. and other tooltips on the top of pages, so that it matches the width of the two columns?

How about if I expand it to half-width, instead of just being a third? That way it will align better with the columns but also keep the text lines shorter and easier to read.

ChengShi-1:
@mccalluc I agree with the half-width, as long as they are aligned, I think it looks good 👍

PR #588, opened 2025-09-10
Add synthetic data generation

### LINKED ISSUE #587: Release contingency tables
On the path to synthetic data.

mccalluc:
- Fix #587


dp_wizard/utils/code_generators/no-tests/_synth_notebook.py
```diff
@@ -0,0 +1,68 @@
...
+IMPORTS_BLOCK
+# -
+
+# Then define some utility functions to handle dataframes and plot results:
+
+# +
+UTILS_BLOCK
+# -
+
+# ### Context
```
ChengShi-1, line 29: should this context be the same hierarchy with prerequisites and results? like ## Context

ChengShi-1 (commented):
i think the UI functionally works well and is straightforward to me. 

As for the UI style, the “Release synthetic data” checkbox now looks somewhat detached and plain. This might be because it doesn’t belong to any particular block, so wrap it with a ui.card or something else may look more integrated. It could be an extra issue for this 
I tried to wrap it with return ui.card( ) locally, as a sample
image

dp_wizard/shiny/results_panel.py
```diff
@@ -136,6 +138,24 @@ def results_requirements_warning_ui():
             ),
         )
 
+    @render.ui
+    def synthetic_data_ui():
+        return [
+            ui.markdown(
+                """
+                Your can select either plain statistics
```
ChengShi-1, line 146: Typo: your can -> you can

dp_wizard/utils/code_generators/no-tests/_synth_notebook.py
mccalluc: No, but thanks for checking: I actually do want it to be an h3 under the h2 for "Analysis". For the stats notebook there are other h3s for each column.

mccalluc: - #597

PR #611, opened 2025-09-24
link to blog posts and other projects

### LINKED ISSUE #610: link privsyn.com
Hi,

I developped a demo website privsyn.com. My vision is the website is easier to use for non-dp experts (since they don't need to install a package etc). I wonder if there's an chance to have that as part of dp-wizard.

mccalluc:
Happy to revise wording, and expand the list, if there are particular projects to point to. 

(Do we even want to have a list like this? I think being able to point at other projects in the same space can inspire some confidence in the user that they've found a good option.)

- fix #610

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,12 +182,33 @@ graph TD
 - These are the only the states that matter. Whether PR is a draft or has assignees does not matter.
 - If we need anything more than this, we should consider a paid plan, so that we have access to more workflows.
 
-## Other resources
+
+## News
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -182,12 +182,33 @@ graph TD
 - These are the only the states that matter. Whether PR is a draft or has assignees does not matter.
```
ChengShi-1, line 182: " These are the only the states ..." there are two "the" here

ChengShi-1 (approved):

PR #628, opened 2025-09-29
Check links

### LINKED ISSUE #547: Check links in generated notebook
We've now got a lot of links to documentation: Great! We should check to make sure they keep on working. Not sure where this scan should happen: In the template source, or in the generated notebook? (Or the HTML render of the notebook?)

mccalluc:
- Fix #547

Might relegate it to a nightly test at some point, but for now there isn't a problem.


ChengShi-1 (commented):
The copilot commented something, they're maybe not strictly required but ok to have.

mccalluc:
@ChengShi-1 , if this needs to be more robust, I can pull in a real HTML parser, but I don't think making the regex more complicated is the right direction. Let me know your preference?

PR #629, opened 2025-09-30
Persist tutorial and dark mode settings

### LINKED ISSUE #619: Persist mode switches across invocations
I get tired of turning off tutorial mode every time.
- On change, write to a git-ignored config file
- On startup, read from file if it's there

mccalluc:
- Fix #619

The `Config` object might be overkill: The bools could just be globals... but this is a good framework to build on if we have any more config options. Let me know what you think?


ChengShi-1 (approved):
I like the current approach, it keeps things organized and is a good foundation for future growth.

Issue #630, opened 2025-09-30
Inline more templates


mccalluc:
Generally easier to work with inline as a function.

PR #632, opened 2025-09-30
Inline query templates

mccalluc:
I was going to work on the queries, but I found bouncing between files confusing. Hope this is an improvement, but it is subjective.
- Towards #630


ChengShi-1 (approved):
Looks good

PR #637, opened 2025-10-01
Identifier truncation

### LINKED ISSUE #347: Use the new group-by-id feature in opendp 0.13
Requires new UI. demo.csv already has student IDs, so that should be smooth.

mccalluc:
- Fix #347 

draft for now.


mccalluc:
Still waiting on OpenDP release. When that is available, it will be better to start fresh, than to try to deal with these conflicts. Closing, but expect to look at these changes again for inspiration.

PR #641, opened 2025-10-02
remove references to google forms

### LINKED ISSUE #639: Remove google forms feedback
For several reasons:
- There hasn't been any outside feedback
- AFAIK, only I am looking at this on a regular basis
- Letting google know every time the application is launched isn't great for software that should be enabling privacy

mccalluc:
- Fix #639

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ be made with private data.
 
 ## Contributions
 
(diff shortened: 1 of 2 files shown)


ChengShi-1 (approved):
Looks good.

PR #642, opened 2025-10-02
date and version in output

### LINKED ISSUE #640: Include date and dp_wizard version number in output
Might help with debugging and user support to know what version is running.

mccalluc:
- Fix #640


ChengShi-1 (approved):
Looks good to me. Copilot mentioned timezone awareness, but since this isn’t a time-sensitive use case, I think it’s fine as is.

Issue #636, opened 2025-10-01
Split panel files into smaller components


mccalluc:
Starting with the `analysis_panel.py`, it should be replaced with something like:

```
analysis_panel/
  __init__.py
  columns_card.py
  grouping_card.py
  budget_card.py
  simulation_card.py
  column_module.py
```

PR #660, opened 2025-10-17
Rearrange UI component files

mccalluc:
- Toward #636
- Actually splitting can follow later.


mccalluc:
When I get around to the package download, I'll be working with the `tmp` directory again, so that would be a good time to clean up the `parent` chain.

PR #664, opened 2025-10-20
Summary of previous choices at top of page

### LINKED ISSUE #649: Usability? Summarize important choices at the top of the tab
With the number of options growing, I find myself on autopilot sometimes and forgetting the exact choices I made earlier in the flow. Would a one-line summary of the previous choices make sense? So
- On Analysis, just include the type of analysis, and the unit of privacy.
- On Download, include that, and a second row with the analysis choices.

See what other people think.

mccalluc:
- Fix #649
- Extends #663

By design, this isn't capturing everything in state: It's a summary. I don't think this needs a surrounding card: We don't need to emphasize it. Let me know if you have any thoughts?

Screenshot 2025-10-20 at 8 06 47 PM


ekraffmiller (approved):
looks good!

Issue #494, opened 2025-06-24
More typing


mccalluc:
With PRs that have been out a long time, I think it will help me with the merges if I have more type information. This is one argument among many.

PR #669, opened 2025-10-22
Start strict type checking

mccalluc:
- Towards #494
- Replace #495

When the pyproject.toml is updated, there are new, fussier warnings in VSCode.
Finding the unused function is a good thing, so it would be nice to push this forward, as time permits.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-exclude = .git,.venv,.venv-*,__pycache__
+exclude = .git,.venv,.venv-*,__pycache__,.local-sessions
 
 # Config recommended by black:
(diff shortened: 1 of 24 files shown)


dp_wizard/utils/argparse_helpers.py
```diff
@@ -1,18 +1,7 @@
...
 from typing import NamedTuple
 
-
-def _existing_csv_type(arg: str) -> Path:
-    path = Path(arg)
-    if not path.exists():
-        raise argparse.ArgumentTypeError(f"No such file: {arg}")
-    if path.suffix != ".csv":
-        raise argparse.ArgumentTypeError(f'Must have ".csv" extension: {arg}')
-    return path
```
mccalluc: Unused! This goes back to when input files could be specified on the command line.

PR #680, opened 2025-10-23
Cleanup tmp directory

### LINKED ISSUE #676: Clarify `tmp/`; Separate session-only files
With zip downloads
- #646 

we need to be sure that each session, even when running locally, has a separate scratch space, and these in turn need to be distinguished from the installation-wide configuration.

mccalluc:
- Fix #676

Q: In the cloud, does the light-dark-mode toggle hit all instances? Or is it bound to session?

A: Confirmed that the config settings are kept within session. They must have separate instances with separate file systems. Since it works, not inclined to worry too much about how it works.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,6 +36,9 @@ jobs:
       - name: Install dev dependencies
         run: pip install -r ${{ matrix.requirements-file }}
 
+      - name: Install pre-commit
+        run: pre-commit install
+
       - name: Install browsers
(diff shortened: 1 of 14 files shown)


mccalluc:
@ekraffmiller, It's two things:
- `tmp` isn't a good name, because we actually want the configuration to stick around.
- `tmp` also conflated two distinct uses: The configurations we actually want to persist, and a scratch space for notebook execution.

Issue #659, opened 2025-10-16
Update the harvard-it-summit-2025 slides


mccalluc:
These are still being referenced, but they are getting out of date.

- Start a new repo that will be a living document (or rename this one?) (or merge into this repo?)
- Fix
  - link in slides
  - Other references to IT Summit: "Are you at the Harvard IT Summit mostly for the free lunch?"
  - Update registry links
  - Add synthetic data to feature list
  - Remove study link

PR #687, opened 2025-10-24
Add slides

### LINKED ISSUE #678: Scale down screenshots
We don't need full size.

mccalluc:
- Towards #659

(Keep open until we can archive the old one.)

- Fix #678

### DIFF
--- a/README-PYPI.md
+++ b/README-PYPI.md
@@ -2,8 +2,11 @@
 
 [![pypi](https://img.shields.io/pypi/v/dp_wizard)](https://pypi.org/project/dp_wizard/)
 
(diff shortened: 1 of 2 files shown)


mccalluc:
@ekraffmiller : huh: the one that comes up as a black square is an SVG. https://www.jetbrains.com/help/pycharm/view-images-in-ide.html
> PyCharm supports all popular image formats, such as PGN, JPG, and GIF.

It would be odd if they left it out, but SVGs are kind of weird. The preview works on VSCode, FWIW.

Updated from main and rerendered the screen shots: Approved?

PR #688, opened 2025-10-27
Download zip: tweak UI

mccalluc:
Following up on the discussion this morning, below the download notebook button we just want to list the files available, styled as links rather than buttons. (If the required information hasn't been filled in, then it's just text.)

Screenshot 2025-10-27 at 3 22 38 PM

### DIFF
--- a/dp_wizard/shiny/panels/results_panel/__init__.py
+++ b/dp_wizard/shiny/panels/results_panel/__init__.py
@@ -104,7 +104,8 @@ def results_server(
     product = state.product
 
     # Analysis choices:
-    # column_names = state.column_names
+    # all_column_names = state.all_column_names
+    # numeric_column_names = state.numeric_column_names
     groups = state.groups
     epsilon = state.epsilon
 
--- a/dp_wizard/shiny/__init__.py
+++ b/dp_wizard/shiny/__init__.py
@@ -16,7 +16,7 @@ from dp_wizard.shiny.panels import (
 from dp_wizard.types import AppState, Product
 from dp_wizard.utils import config
 from dp_wizard.utils.argparse_helpers import CLIInfo
-from dp_wizard.utils.csv_helper import read_csv_names
+from dp_wizard.utils.csv_helper import read_csv_names, read_csv_numeric_names
 
 _shiny_root = package_root / "shiny"
 _assets_root = _shiny_root / "assets"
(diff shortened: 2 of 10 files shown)


dp_wizard/shiny/panels/results_panel/__init__.py
```diff
@@ -178,73 +182,60 @@ def clean_download_stem() -> str:
...
-        ]
+        )
 
     @render.ui
     def download_code_ui():
         disabled = not weights()
-        return [
-            ui.h3("Download Code"),
+        return ui.card(
+            ui.card_header("Code"),
```
mccalluc: Using cards is more consistent with the UI in the preceding tabs.

dp_wizard/shiny/panels/results_panel/__init__.py
```diff
@@ -255,43 +246,13 @@ def download_code_ui():
                     but does not contain any data or analysis results.
                     """
                 ),
+                responsive=False,
```
mccalluc: Since this is now a two column UI, we want this box to span the full width of the card.

dp_wizard/shiny/panels/results_panel/download_options.py
```diff
@@ -26,7 +25,6 @@ def clean_description_md(self) -> str:
             ".zip",
             "folder-open",
             "A zip file with results and code for a reproducible analysis.",
-            None,
```
mccalluc: Just weren't using the optional message kwarg. Drop it.

dp_wizard/shiny/panels/results_panel/download_options.py
```diff
@@ -104,35 +91,80 @@ def clean_description_md(self) -> str:
...
+def table_of_contents_md():
+    """
+    >>> print(table_of_contents_md())
+    - README.txt
+    ...
+    - Table (.csv): The same information, but condensed into a CSV.
+    """
+    included_names = ["Notebook", "HTML", "Script", "Report", "Table"]
+    included_options = [_download_options[name] for name in included_names]
+    return "- README.txt\n" + "\n".join(
```
mccalluc: This changes in the next PR: The README gets a description like the others.

dp_wizard/shiny/panels/results_panel/download_options.py
```diff
@@ -104,35 +91,80 @@ def clean_description_md(self) -> str:
...
+    disabled: bool,
+    button: bool,
+):  # pragma: no cover
+    opt = _download_options[opt_name]
     clean_name = re.sub(r"\W+", " ", opt.name).strip().replace(" ", "_").lower()
     kwargs = {
-        "id": f"download_{clean_name}",
-        "label": f"Download {opt.name} ({opt.ext})",
+        "id": f"download_{clean_name}_{'button' if button else 'link'}",
+        "label": f"{'Download ' if button else ''}{opt.name} ({opt.ext})",
```
mccalluc: Buttons should be imperatives; Links should just be noun phrases.

mccalluc:
@ekraffmiller , For now I'd like to keep the second column:
- Only one of the items in the column on the right is actually part of the zip bundle
  - Adding all of them feels like it would clutter the zip bundle, so I don't want to do that.
  - But I don't want to just drop them: They could be useful
- That second column is what is shown in cloud mode, and I'd kind of like to avoid splitting the UI experiences too far apart between the two modes.

But I'm not super happy with this either. Filed an issue to revisit this:
- #715

ok to approve? Thanks!

Issue #666, opened 2025-10-21
Put all columns, used or not, into analysis plan


mccalluc:
I think there's a request to describe the whole CSV. Capture all columns, even if not used in analysis.

PR #691, opened 2025-10-28
List columns in readme

mccalluc:
- Towards #666

For a codebook notebook, we'd need this in the analysis plan, but this is a good step on its own.


ChengShi-1 (approved):
Good! It's pretty nice to have the readme downloadable

PR #697, opened 2025-10-28
Limit column selection to numerics

### LINKED ISSUE #694: Limit column selection to numeric columns
We have a schema from polars, but we're just using the column names. We could limit the column choices to numeric columns. (Grouping should still be available for all columns.)

mccalluc:
- Fix #694 
- Groups can still be selected from all columns
- When running in cloud, we assume all columns could be numeric. Right now, I don't think it would be worth the added UI complexity and work for the user to add this.

### DIFF
--- a/dp_wizard/shiny/panels/analysis_panel/column_module.py
+++ b/dp_wizard/shiny/panels/analysis_panel/column_module.py
@@ -260,14 +260,17 @@ def column_server(
             )
 
         def upper_bound_input():
-            return (
-                ui.input_text(
-                    "upper_bound",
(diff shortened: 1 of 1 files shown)


ChengShi-1 (approved):
Found an ui bug when I test, but it isn't relevant to this PR. https://github.com/opendp/dp-wizard/issues/713

mccalluc:
Playwright timeout: `waiting for get_by_text("Download Report (.txt)")`. Probably a race condition? Rerunning test. That UI is changing soon, so worry about it later, if it persists.

PR #704, opened 2025-10-29
For synthetic data, don't show an analysis selector

### LINKED ISSUE #695: For synthetic data column selection, there shouldn't be an analysis selection
It should still work if the user returns to the first tab and changes the product: internally, maybe it's a histogram internally when the mode is synthetic data?

### LINKED ISSUE #703: Drop count
Doesn't really make sense on one column, unless there are more options to select a subset.

Also, see:
- #643

That will do counts the right way.

mccalluc:
We only need bins. The preview visualization is still useful for building intuition, I think.

(If you were to select mean or median, and then go back to the first tab, and change to synthetic data, the types are now locked: This PR is just a surface UI change. Could imagine doing it at a deeper level, but this seems like enough.)

Also drop count: it was the odd one out, and it really should be done across the whole CSV.

- Fix #695
- Fix #703

### DIFF
--- a/dp_wizard/shiny/panels/analysis_panel/column_module.py
+++ b/dp_wizard/shiny/panels/analysis_panel/column_module.py
@@ -315,7 +315,12 @@ def column_server(
                 width=label_width,
             )
 
-        analysis_name = input.analysis_type()
+        # Preserve the user's choices behind the scenes,
+        # but only show mean and median column UI if actually calculating stats:
+        # otherwise show the histogram UI.
+        analysis_name = (
+            input.analysis_type() if product() == Product.STATISTICS else histogram.name
+        )
 
         # Had trouble with locals() inside comprehension in Python 3.10.
         # Not sure if this is the exact issue:


mccalluc:
Good point: Even if they had selected histogram, the code sample isn't what is used for synthetic data. And the note about "Since this stat is just a single number" doesn't make that much sense either.

Stepping back: If we're in synthetic data mode, we really _only_ do histograms. So the challenge is what do if the user is switching between tabs, and had selected mean or median.
- Either dropping them from display, or dropping them from the query plan entirely, would probably be confusing ("I selected `col_x`; where did it go?"), and might be hard to implement, regardless. 
- Right now in this PR, the only part of the display that changes depending on mode is the first row; I should see if it's feasible to have the lower part also depend on mode, so if we're in synthetic data mode, it would show the histogram ui, regardless of what stat is chosen.  (Just as downstream for synthetic data, the choice of stat is ignored, and we're doing histograms under the hood.) 

If I can make that work, I should still make sure the whole ignore-users-stat-selection-if-synthetic is explained in the source code. If it's confusing now, it's only going to be more confusing in the future.

Thanks! In my court.

ekraffmiller (approved):
looks good!

PR #731, opened 2025-11-19
Tidy up results page UI

### LINKED ISSUE #729: UI tweak: Matching panel heights on results page?
This is fussy, but it would be a little bit tidier if the panel heights matched on the results tab.

mccalluc:
- Fix #729
- Add icons, since all the other cards have icons
- Document another Shiny quirk

For reviewer:
- Does the file-with-a-check-mark seem like a good way to suggest that these files are special?
Screenshot 2025-11-19 at 11 35 05 AM


ekraffmiller (approved):
I think it looks good, the icons are unobtrusive but help with the visual organization of the page.

PR #737, opened 2025-11-21
Public grouping keys

### LINKED ISSUE #736: Prompt for members of groups
If the user knows the members, great! We should use the information we have to be efficient with the budget.

Discussed this with the IQSS folks earlier, and adding a panel for each group seems ok.

mccalluc:
- Fix #736

This gets a little complicated because the user provided values need to match the type of the column, so I've added the inferred schema to the appstate.

For reviewer:
- Does the UI seem ok? Does the explanation make sense?
- Does the generated notebook seem good? Is the code idiomatic?
- QA: Does it actually seem to work? Have I added enough test coverage?

TODO:
- #742


ekraffmiller:
> @ekraffmiller , It should work with integer columns (and I added an integer `class_year` to the sample.csv, and that worked locally), but it obviously doesn't work for you. Looking at the stack trace, it looks like the type sniffing failed: `"0", "1"` are strings rather than integers.
> 
> Can you attach the CSV you're using, either here, or in Slack? And are you just grouping, or are you grouping and specifying public keys? (Which reminds me that the public keys should be included in the summary line at the top.) Thanks!

Hi @mccalluc , here is the file I'm using. I only get the error when I specify the public keys:
[pums_1000.csv](https://github.com/user-attachments/files/23883739/pums_1000.csv)

ekraffmiller (approved):

PR #739, opened 2025-11-24
Misc: Weight enum, testing improvements

mccalluc:
- From #725

The counts PR should be redone from scratch, since we decided it would be better to always do a DP count. That said, there are several other good changes in that PR that should be kept:
- Weights enum strengthens our type signatures
- Subsetting tests helps them run in a reasonable period of time.
- Include a full path to the CSV fixture in generated code, so it can be run out of the box.


ekraffmiller (approved):
looks good

PR #753, opened 2025-12-10
Check that example code in slides is up to date

### LINKED ISSUE #706: Testing: Confirm that code samples in slides are up to date
Possibly download the source for the notebook, and then make sure that each codeblock in the index.md also is in the source?

### LINKED ISSUE #708: Confirm that slides are up to date
- distinct from #707

Two possibilities:
- run pandoc during tests, and make sure index.html is the same.
- move the pandoc call into ci, and don't check in index.html at all.

mccalluc:
- Fix #706
- Fix #708

A different approach would be to move the pandoc build entirely into an action, instead of checking in index.html. My sense is that it is useful to be able to build and view it locally... and if we're doing that much, then might as well check it in, rather than trying to debug code that only runs in github actions.


docs/index.html
```diff
@@ -737,9 +737,9 @@ <h3 id="opendp-demo">OpenDP Demo</h3>
 <h4 id="prerequisites">Prerequisites</h4>
```
mccalluc: `index.html` is generated from `index.md`: Doesn't need review.

PR #755, opened 2025-12-11
Warn about bad CSVs

### LINKED ISSUE #752: Warn about bad CSVs
- Repeated column names
- missing columns names
- unusual column names (maybe it doesn't have a header row?)
- Or maybe not even a csv? (Perhaps tab delimited, for example)

> I think this looks good, and agree that we don't necessarily have to cover every edge case in the column names, but it made me think that maybe we could have an error displayed to the user if the .csv is malformed in some way?  For example, if there are multiple columns with the same name, or for any other reason.  The app allows the user to upload any file that is .csv, but the file extension doesn't guarantee that the file is well formed.  That could be addressed in another PR though.

_Originally posted by @ekraffmiller in https://github.com/opendp/dp-wizard/pull/745#pullrequestreview-3562969628_

mccalluc:
- Fix #752

For reviewer:
- The playwright tests haven't been extended, but the unit testing is solid: Good enough?
- Looking at `test_csv_info`, does the distinction between warnings (which are displayed in the UI, but don't block progress) and errors (which do block until they are resolved) make sense? Should the dividing line be moved one way or the other?
- Ideas for other kinds of errors to catch?
- We already check that the public and private columns match... in the interests of simplicity, not checking that the schemas match. Filed #757. Ok to punt?

### DIFF
--- a/dp_wizard/types.py
+++ b/dp_wizard/types.py
@@ -175,7 +175,7 @@ class CsvInfo:
 
     def get_schema(self) -> dict[ColumnName, pl.DataType]:
         if self._errors:
-            return {}
+            return {}  # pragma: no cover
         return self._schema
 
     def get_all_column_names(self) -> list[ColumnName]:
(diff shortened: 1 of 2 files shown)


mccalluc:
This got complicated because letting users specify grouping keys came in at the same time, and that added the schema as part of state. The resolution is to just keep the csv_info in state, and everything else can be derived from it.

(On cloud, use the user input to construct a temp CSV for inference.)

PR #759, opened 2025-12-15
Contingency table download

### LINKED ISSUE #758: If synthetic data, make contingency table available for download
The contingency table is the real product of the DP release, and in some ways is more informative than the synthetic data: For example, the bins for numeric fields are explicit.

mccalluc:
- Fix #758

For reviewer:
- Does the explanatory text seem ok?
- Does the error behavior seem ok? (Could imagine having a warning earlier, if the the user has selected a whole lot of groups, but probably not worth overdoing: Maybe no one even hits this error.)

### DIFF
--- a/dp_wizard/utils/code_generators/abstract_generator.py
+++ b/dp_wizard/utils/code_generators/abstract_generator.py
@@ -345,10 +345,12 @@ are ignored because of errors, it will bias results.
             from math import prod
 
             possible_rows = prod([len(v) for v in contingency_table.keys.values()])
+            max_rows = 100_000
             contingency_table_melted = (
                 contingency_table.project_melted(COLUMNS)
-                if possible_rows < 100_000
-                else "Too big!"
+                if possible_rows < max_rows
+                else f"Contingency table could be more than {max_rows} rows; "
+                "Consider querying for just the information you need."
             )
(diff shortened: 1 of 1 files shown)


dp_wizard/utils/code_generators/abstract_generator.py
```diff
@@ -341,11 +341,12 @@ def template(synth_context, COLUMNS, CUTS, KEYS):
             from math import prod
 
             possible_rows = prod([len(v) for v in contingency_table.keys.values()])
-            (
-                contingency_table.project_melted([COLUMNS])
+            contingency_table_melted = (
+                contingency_table.project_melted(COLUMNS)
                 if possible_rows < 100_000
                 else "Too big!"
```
ekraffmiller: Can this error provide more guidance to the user?  As a simple addition, it could say that the max number of rows is 100,000.

ekraffmiller (commented):
Looks good, I just have a comment about the error message.

mccalluc:
Longer message:
```
else f"Contingency table could be more than {max_rows} rows; "
"Consider querying for just the information you need."
```
... but this is in the generated code, so the focus should stay on a simple demonstration of the library, and if that proves difficult, we might consider a fix upstream to make the library more usable for anyone.

PR #766, opened 2025-12-17
Exponential histogram for finding bounds

### LINKED ISSUE #764: Bounds helper
- See also #586

Just prompt for an extreme value, and it will make a exponential histogram.

mccalluc:
- Fix #764

For reviewer:
- Is "Bounds" a good name? It's a description of what it's used for, rather than what it is, but I thought this is better than exponential histogram.  (Concise is good!)
- Is this a good interface? Could imagine also asking for min, or a bin count (or a base for the exponent)... but it seems like most of the time this is plenty.
- Does the in-app explanation make sense?
- Ok that I'm continuing to ignore positive values much less than 1? They do matter in other disciplines, but my sense is that for human subjects data small positives aren't that common.
- I don't think I want to lengthen the descriptions on the other stats, but I could be wrong about that.


mccalluc:
- Ellen: asking for "upper bound" doesn't make sense if this is about finding bounds
- Let's not prompt the user: hard code 1,000,000,000!

PR #772, opened 2025-12-18
use template `when` feature

### LINKED ISSUE #771: Upgrade dp-wizard-template, and use `unless`

mccalluc:
- Fix #771 

For reviewer:
- This is really about the upstream library: Is the behavior of the `unless` kwarg intuitive? I had started with `if`, but that's a python keyword, so it doesn't work, but I could try other words, if you have strong feelings.


mccalluc:
@ChengShi-1 , would a `when` kwarg be better? and drop `unless`, so there's one right way to do things?

PR #777, opened 2025-12-22
Elevate to yellow for warning

### LINKED ISSUE #770: Red instead of blue for error messages?
>> would it be possible to change the color of the error message UI so it stands out from the other help text on the page?
>
>I could, but
>- There are other error messages in the app, and if this changes I'd like to handle them all at once to be consistent.
>- Should both warnings and errors get the same treatment here? I'm not sure. (The little message at the bottom of the page that explains why the button isn't enabled is a kind of very-low-priority warning message.)
>- I had tried red boxes earlier, and it just seemed a bit harsh. If we do want to distinguish this, maybe a favicon? or text styling?

_Originally posted by @mccalluc in https://github.com/opendp/dp-wizard/issues/755#issuecomment-3672113473_

mccalluc:
- Fix #770

I'm not sure about this, but it might be that I'm just used to the status quo.

- While these warnings are legitimate, things can't go very wrong with DP Wizard. We don't need to worry about deleting user data or anything like that, so we may not need high-level alerts.
- In some cases, like "Maximum row count should be at least 100", we show a warning before the user has even done anything. I don't think we want to make it seem like the user has done something wrong before they even start, but at the same time, I don't think we want to have a mix of colors.
- If we do change the color, we should probably do it everywhere.
- I don't think there is a danger of these being overlooked, so I don't think they need to be highlighted.

All that said, I won't block if this seems like the right direction to other folks.

(Note that the hidden info boxes should stay blue or grey, regardless.)

Screenshot 2025-12-22 at 5 48 40 PM

### DIFF
--- a/dp_wizard/shiny/panels/dataset_panel/__init__.py
+++ b/dp_wizard/shiny/panels/dataset_panel/__init__.py
@@ -339,9 +339,11 @@ def dataset_server(
 
     @reactive.effect
     def set_is_dataset_selected():
+        info = csv_info()
         is_dataset_selected.set(
             contributions_valid()
-            and not csv_info().get_is_error()
+            and not info.get_is_error()
+            and len(info.get_all_column_names()) > 0
             and not get_row_count_errors(max_rows())
             and (in_cloud or not csv_column_mismatch_calc())
         )


mccalluc:
How about if an empty CSV is not a CSV validation error? Validating that there are columns (either from the CSV, or from the user in cloud mode) can be a separate validation.

ekraffmiller (approved):
looks good

PR #792, opened 2026-01-08
Tweak messages if synthetic data

### LINKED ISSUE #681: "Select columns to calculate statistics on": Should be different for synth data
Tweak message

Also: "Select columns to group by, or leave empty to calculate statistics across the entire dataset"

mccalluc:
- Fix #681
- Towards #636 : I realized that these can't be pulled out completely because the selectize lists won't update in dynamic ui.

For reviewer:
- Do the messages look ok? We don't really have "grouping" for synthetic data, but changing that based on the product type would get really annoying, and this might be fine with users.


ekraffmiller:
@mccalluc if 'grouping' isn't appropriate for synthetic data, can the input in the Grouping card be disabled, when the user chooses synthetic data?

ekraffmiller:
> > can the input in the Grouping card be disabled, when the user chooses synthetic data?
> 
> It's still useful for generating synthetic data, because the string-valued columns can't be selected for analysis... Maybe that's what needs to be tackled first? But I don't want to make the code much more complicated for something that feels like a pretty minor UI feature.
> 
> Not sure about the best course here: I'll give it more thought, but if you have suggestions, that's great, too.

I'm not sure I understand your original comment - 'We don't really have "grouping" for synthetic data, but changing that based on the product type would get really annoying, and this might be fine with users.'

mccalluc:
What is now called "grouping" is still very useful for synthetic data, even if it's not actually grouping:
- Non-numeric fields aren't available for "columns"
- Only "grouping" lets us specify known values

It wouldn't be impossible for these to change, but it would get messy. Right now an analysis plan can be used either for synthetic data or for stats, and the user can toggle between the two on the first tab, without invalidating anything on the second tab.

For that reason, I think it makes sense to confine the search space to changes of wording, particular changes of wording just for synthetic data.

When synthetic data is selected, instead of `Columns / Grouping`, would `Numeric / Categorical` do?

dp_wizard/shiny/panels/analysis_panel/__init__.py
```diff
@@ -109,6 +56,75 @@ def analysis_ui():
     )
 
 
+def _columns_card_ui():
+    return (
+        ui.card(
+            ui.card_header(columns_icon, "Columns"),
+            ui.output_ui("select_columns_message_ui"),
```
eddiestudies, line 63: Can these be a constant class or is it required to be a string? I think they get pulled into the function render.ui function but it might help to group the keys into
ColumnUIMessages.SELECT

It helps avoid typo bugs and groups the keys. It might not be possible so feel free to let me know if that is the case I am not familiar with the render.ui infrastructure.

dp_wizard/shiny/panels/analysis_panel/__init__.py
```diff
@@ -109,6 +56,75 @@ def analysis_ui():
...
+    )
+
+
+def _privacy_card_ui():
+    return (
+        ui.card(
+            ui.card_header(budget_icon, "Privacy Budget"),
+            ui.markdown(
+                f"""
+                What is your privacy budget for this release?
```
eddiestudies, line 97: I think this ui.markdown should be consistent with the ui.output_ui changes above. Mostly to keep to one convention so that future changes can do the same. 

I do not think moving all of the keys into constants makes sense although originally I thought it would. This seems like an artifact of using the ui library and the alternative makes the dev experience worse.

For context, I was thinking of suggesting moving "select_groups_message_ui" into UIMessages.GROUPS... but the function matches that string exactly, and adding a constant would require an extra search to find the correct caller.

I tend to avoid using strings as direct keys unless absolutely necessary but in this case I don't see a better option

dp_wizard/shiny/panels/analysis_panel/__init__.py
mccalluc: It's one of the parts of the framework which annoys me. I don't want to introduce a new constant, but I think I could reference the actual function:
```
ui.output_ui(select_columns_message_ui.__name__),
```

dp_wizard/shiny/panels/analysis_panel/__init__.py
mccalluc: > I think this ui.markdown should be consistent with the ui.output_ui changes above. 

Do you mean `select_columns_message_ui` or `select_groups_message_ui`? Both of those have conditional logic that may have required pulling it out, but with something that is just a single short markdown block, I'd prefer to keep it in line. 

(Another frustration I have with Shiny is that refactorings which shouldn't make a display difference sometimes do. I don't remember the details, but it was something like calling it through another function added a div and the css flex styling broke. Or something like that!)

eddiestudies: Ok. So the convention is if there is logic then wrap, otherwise inline it. Thanks for clarifying why it is handled differently

eddiestudies (approved):
lgtm

PR #795, opened 2026-01-12
Random tests, and better input validation

### LINKED ISSUE #793: Try the hypothesis test library

mccalluc:
- Fix #793
- Add more user input checks. (Guided by bugs that hypothesis identified.)

For the reviewer:
- I started a new test file for this: While the structure is similar to other tests, I wanted to make sure I didn't confuse the two... but if this doesn't seem right, I could combine.
- Is the text of new error messages readable?

So that there's not a moving target, I'm going to make a branch from this for the next step. If there a non-blocking suggestions here, I'd prefer to handle them in the next PR, just to keep the merge process simple.


ekraffmiller (approved):
looks good

PR #824, opened 2026-01-26
Add FAQ

### LINKED ISSUE #822: FAQ
Starting with whether DP is anonymization.
- Should be driven by questions from user: Let's not try to guess their questions.
- Show in app, and link from readme.

mccalluc:
- Fix #822
- Reference FAQ inside the application: replace the "About" tab
- Rephrase the pre-fill form, so it's more clear what it does. (We don't want people without a github account to waste their time.)
- Link to FAQ from README. (Link won't work until after merge.)

For reviewer:
- Proof read!
- Would it be good to get a dp-wizard email? Would anyone besides me like to receive anything that is incoming?

### DIFF
--- a/README-PYPI.md
+++ b/README-PYPI.md
@@ -6,14 +6,14 @@ DP Wizard makes it easier to get started with differential privacy,
 the addition of calibrated noise to aggregate statistics to protect the privacy of individuals.
 DP Wizard demonstrates how to calculate DP statistics or create a synthetic dataset from the data you provide.
 
-You can run DP Wizard locally and upload your own CSV,
-or use the [cloud deployment](https://mccalluc-dp-wizard.share.connect.posit.cloud/) and only provide column names to protect your private data.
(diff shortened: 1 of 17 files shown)


README-PYPI.md
```diff
@@ -15,6 +15,8 @@ Options for running DP Wizard:
 - Install from [PyPI](https://pypi.org/project/dp-wizard/): `pip install 'dp-wizard[app]'; dp-wizard`
 - Install from [source](https://github.com/opendp/dp-wizard): See developer instructions.
 
+See the [FAQ](https://github.com/opendp/dp-wizard/blob/main/dp_wizard/FAQ.md) for more information.
```
ekraffmiller:
```suggestion
See the [FAQ](dp_wizard/FAQ.md) for more information.
```

ekraffmiller: Ok, sorry about that I didn't think of the context of how PYPI is being used. In our docs we use relative urls to avoid this problem of branches in the url.

PR #825, opened 2026-01-26
Remove text report; replace with YAML config download

### LINKED ISSUE #821: yaml export
With 
- #815 

there's a human-readable summary of the analysis in the README, so we don't need to report that information in the report.txt.

What _could_ be useful would be an unambiguously machine-readable export, that could also be loaded at startup to set parameters.

(Looking at the code now, the report hasn't kept up: it's missing grouping information.)

mccalluc:
- Fix #821
- With #815 in place, having a second human-readable summary of the analysis doesn't make sense: Instead, provide an export which unequivocally machine-readable.
- The text report hadn't kept up with the all the options available in an AnalysisPlan: better to export that directly.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,34 @@
+# Include any files or directories that you don't want to be copied to your
+# container here (e.g., local build artifacts, temporary files, etc.).
+#
+# For more help, visit the .dockerignore file reference guide at
+# https://docs.docker.com/engine/reference/builder/#dockerignore-file
+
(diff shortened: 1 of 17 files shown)


ekraffmiller (commented):
I think this looks ok, but is it necessary to merge this before an import feature is ready?  In https://github.com/opendp/dp-wizard/issues/826 you are asking if the import is needed, so maybe that should be discussed first?

PR #844, opened 2026-02-06
Split playwright downloads into separate tests

### LINKED ISSUE #837: 3.13, requirements-dev: test_local_app_downloads often fails
That particular combination in the matrix seems particularly fragile. [for example](https://github.com/opendp/dp-wizard/actions/runs/21614503906/job/62290231952):
```
FAILED tests/test_app.py::test_local_app_downloads[chromium] - playwright._impl._errors.TimeoutError: Timeout 30000ms exceeded while waiting for event "download"
```
- Make a venv for that combo locally and try to reproduce?
- Add more logging: is a particular download the problem?
- Split up the test: Maybe download the scripts in a separate test?
- Last resort, add an automatic retry?

mccalluc:
- Use pytest parameters, so each download option gets its own test run.
- Possibly fix #837
- This is a weaker, slower test: If there was an interaction between successive downloads that caused problems, we'd miss it now... but just rerunning the tests that fail wasn't moving us towards a solution, either.

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -339,16 +339,16 @@ def test_local_app_downloads(
     stem_locator.fill(new_stem)
     expect(stem_locator).to_have_value(new_stem)
 
-    download_option.requires_release
+    warning_expected = download_option.requires_release
 
     # -- Define Analysis --
     page.get_by_role("tab", name="Define Analysis").click()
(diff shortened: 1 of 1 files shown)


rtreacy (approved):
Looks good. Discussed this morning bi-weekly meeting

PR #877, opened 2026-02-25
"unit of privacy" -> "unit of protection"

### LINKED ISSUE #857: "Unit of Privacy" -> "Unit of Protection"
Less jargon.

mccalluc:
- Fix #857

This language change was suggested by the DP experts.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -909,7 +909,7 @@ understand distributions.</p>
 data</h3>
 <p>Text or image data can’t just be dropped into DP.</p>
 <p>Even if it can be reduced to a feature vector, can you define a unit
-of privacy?</p>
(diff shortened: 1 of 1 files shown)


dp_wizard/shiny/panels/dataset_panel/__init__.py
```diff
@@ -351,7 +351,6 @@ def input_contributions_ui():
             ui.markdown(
                 f"""
                 How many **rows** of the CSV can {entity_phrase} contribute to?
-                This is the "unit of privacy" which will be protected.
```
mccalluc: This is now redundant: With a better name, we don't need to explain what the name means.

PR #881, opened 2026-02-26
Add new low-interaction summary analysis

### LINKED ISSUE #868: Low interaction summary: All the columns!
(Might be dupe? but can't find the other.)

CSVs can have hundreds of columns, and data curators don't have a particular analysis in mind: They just want to get some summary out.
- New output mode to complement stats and synthetic data
- Get all the columns:
  - For strings, try to capture distinct values.
  - For numbers, 
    - order of magnitude histogram. (See #766, though that would be selected, by the user, and we just want to apply it everywhere.)
    - or 5% and 95% percentiles, with exponential candidates 👈 I think this is the better option!
- Only parameter choices are
  - upper bound on rows
  - contributions
  - epsilon

mccalluc:
WIP
- Fix #868

### DIFF
--- a/dp_wizard/shiny/panels/dataset_panel/__init__.py
+++ b/dp_wizard/shiny/panels/dataset_panel/__init__.py
@@ -543,6 +543,9 @@ Choose both **Private CSV** and **Public CSV** {PUBLIC_PRIVATE_TEXT}
                   selected columns, and the correlations between columns.
                   This is less accurate than calculating the desired
                   statistics directly, but can be easier to work with downstream.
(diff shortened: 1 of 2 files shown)


mccalluc:
Proposed analysis doesn't produce good results.
- See https://github.com/opendp/opendp/issues/2644

Blocked until we have something that produces better results.

mccalluc:
On hold. Going to get more feedback about what kind of analysis is needed, and how much accuracy is required. See https://github.com/mccalluc/no-dp-wizard

PR #885, opened 2026-02-27
remove invariants, add dp.len

### LINKED ISSUE #371: Try to get rid of invariants
Mike's note during 0.3 release review:
> not a big fan of having preset margins but I know you're working on reducing these assumptions. Looks like you added a delta, but aren't using it. Since delta is there, the second "keys" margin can be completely dropped. I guess it's there because mean and quantiles might want it. The lib could be more forgiving and detect that mean computes lengths if the input expression is row-by-row, so that counts don't have to be released twice.

Try to get rid of invariants! Not what we want on an introductory tool: Only should resort to them when the results are poor and you can compromise the privacy guarantee.

- When releasing a histogram we don't need the margin, but keep the delta
- For the mean, needs either margin OR another query for length and keep delta
- For quantiles, same as mean

This will probably require more conditional logic in the code generation: Maybe clean that up before starting this? (ie, pull code in stand-alone files into functions.)

Mike adds:
> actually, when you release means, you could generate:
> ```
> context.query().group_by(...).agg(dp.len(), pl.col("A").dp.sum(...)).release().collect().with_columns(pl.col("A") / pl.col("len"))
> ```
> This way you get counts for margins for free. The task in the OpenDP lib would be to rewrite queries in this way internally

mccalluc:
- Fix #371
- Disable medians, at least until https://github.com/opendp/opendp/issues/2640 is resolved.

Since this is a substantial change in the analysis, manually exercising it would be good.


ekraffmiller:
@mccalluc can you provide more info on the manual testing needed for this PR?

PR #887, opened 2026-03-05
Light blue background on the tutorial cells in HTML export

### LINKED ISSUE #886: use dp-wizard-templates 0.9.0, and style notebook cells
For the Tutorial, pick a light blue that matches the application.

mccalluc:
Also hide the blob of json in the first cell of ipynb exports.

- Fix #886


ChengShi-1 (approved):
It looks good with the light blue background in html export file.
Image

PR #888, opened 2026-03-05
more doctests for make_cut_points

### LINKED ISSUE #866: Handle bin widths less than 1
I think the bin helper only returns integers right now... but I have seen social science data between -1 and 1!

mccalluc:
- Close #866, although it looks like it already works as desired.

I think I may have remembered a time before `round_2` when I was just using `int`.

I'd be happy to just close this, if you don't think the tests add value, for us, and for people looking at the generated notebooks.


ekraffmiller (approved):
always good to add more tests! :)

Issue #863, opened 2026-02-20
Latency is too high for epsilon slider


mccalluc:
With sliders, users expect quick response, and that's definitely not us.
- Make it faster
  - add timing: is it the render? or the DP?
  - if render, consider other library
  - if DP, Mike thinks the CSV parsing could be the slow bit: consider using parquet?
- Or, use a different UI... but better to just make it fast!

Update: IQSS folks didn't think this is critical right now, at least with small numbers of columns.

PR #889, opened 2026-03-05
Let estimated rows work with public data

### LINKED ISSUE #861: Estimated rows isn't used with public data
Vague memory that the known row count was added to the select list; Or we could just hide that section if we know the number of rows? In any case, right now the select is shown, but changing it doesn't have any effect, apart from prompting a re-render.

mccalluc:
- Fix #861
- Simplify UI code in `analysis_panel/__init__.py`: Minimize the code inside the if-else, and move the return outside.
- Use polars `sample` to generate a Lazyframe of the requested size, either smaller or larger than the original. (lazyframe -> dataframe -> lazyframe  is kludgy, but I don't have a better alternative.)
- Inside `make_accuracy_histogram`, rename `row_count` parameter to `max_length`, since that's what it is used for.

For reviewer:
- Adding more computation will slow down the interface a little bit. With a thousand line file I don't notice a difference, but is this the direction we want to go in? (See #863)
- The only change in the UI is in the plot, so I don't have good idea for an automated test. Let me know if you have ideas?

### DIFF
--- a/dp_wizard/shiny/panels/analysis_panel/column_module.py
+++ b/dp_wizard/shiny/panels/analysis_panel/column_module.py
@@ -18,11 +18,10 @@ from dp_wizard.shiny.components.outputs import (
 )
 from dp_wizard.types import ColumnName, Product, StatisticName, Weight
 from dp_wizard.utils.code_generators import make_column_config_block
-from dp_wizard.utils.code_generators.analyses import (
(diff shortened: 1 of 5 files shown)


mccalluc:
@ekraffmiller : Let me move this back to draft: I think the sampling does fill a gap, but with it probably being O(n^2), it's not something to merge now. Thanks for catching this.

PR #897, opened 2026-03-06
delta=0 if groups

### LINKED ISSUE #883: If we have enumerations for all groups, delta can be zero

mccalluc:
- Fix #883

@Shoeboxam , since this is about demonstrating correct API usage, could you look at this PR?

### DIFF
--- a/dp_wizard/utils/code_generators/abstract_generator.py
+++ b/dp_wizard/utils/code_generators/abstract_generator.py
@@ -11,7 +11,6 @@ from dp_wizard.utils.code_generators import (
     make_privacy_loss_block,
     make_privacy_unit_block,
 )
-from dp_wizard.utils.code_generators.analyses import histogram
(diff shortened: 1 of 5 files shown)


dp_wizard/utils/code_generators/abstract_generator.py
```diff
@@ -249,30 +249,23 @@ def _make_partial_stats_context(self):
...
-
-        is_just_histograms = all(
-            plan_column[0].statistic_name == histogram.name
-            for plan_column in self.analysis_plan.analysis_columns.values()
-        )
-        margins_list = (
-            # Histograms don't need margins.
-            "[]"
-            if is_just_histograms
-            else self._make_margins_list(
```
mccalluc: `is_just_histograms` wasn't a good check, because we actually do need the margins, albeit with public keys, if we want to be able to set delta to zero.

ekraffmiller (approved):
looks good

PR #915, opened 2026-03-11
"CSV" to "CSV or TSV" or "data source"

mccalluc:
- Missed these during #879
- Fix #913
- Tried to use "or TSV" when it's talking about the particular options available, and "data source" when it's just talking about the earlier part of the UI... but if the phrasing anywhere seems awkward, please comment!
- Retained "CSV" when talking about outputs, and in the generated code, where the user TSV has already been converted.
- FYI: Filed #914, but not planning to act on that without user input.

I'm using "TSV" to suggest _any_ tab-delim file: The tutorial box explains:
> Internally, OpenDP works best with CSVs; Other formats will be converted to CSV. Any of these extensions are accepted: `.csv`, `.tsv`, `.tab`.

### DIFF
--- a/README-PYPI.md
+++ b/README-PYPI.md
@@ -51,7 +51,7 @@ options:
   --no_browser  By default, a browser is started; Enable this for no browser.
   --reload      Enable to watch source directory and reload on changes.
 
-Unless you have set "--demo", you will specify a CSV inside the application.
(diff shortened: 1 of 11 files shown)


ekraffmiller (approved):
looks good

PR #917, opened 2026-03-11
add "Tutorial" label

### LINKED ISSUE #855: Label toggle with "Tutorial"
~~Someone might not want full tutorial mode, but they might need a little hint. Also, the affordance of the little circles isn't clear: Right now, you can hover or click and nothing happens -- User needs to connect it to the toggle up in the corner.~~

It may be sufficient just to put a clear label on the toggle. I was too worried about screen real estate.

mccalluc:
- Fix #855

Screenshot 2026-03-11 at 3 36 04 PM

It's higher than I'd like, but I spent some time wrangling the CSS, and although I was able to bring the controls down, moving the "Tutorial" down with `margin-top` also lowered the horizontal grey line, and separated it from the tab outline.

If anyone else wants to try, feel free! If this looks too ugly, don't feel obliged to approve.

There's already a little custom CSS in styles.css for this part of the UI:
```
/*
The tutorial mode toggle in the navbar is positioned slightly too high.
*/
.shiny-input-container:has(#tutorial_mode) {
    margin-top: 0.15em;
    margin-bottom: -0.15em;
}
```

### DIFF
--- a/dp_wizard/shiny/__init__.py
+++ b/dp_wizard/shiny/__init__.py
@@ -67,15 +67,14 @@ def _make_app_ui(cli_info: CLIInfo) -> Tag:
                 ui.input_switch(
                     "tutorial_mode",
                     ui.tooltip(
-                        tutorial_icon,
+                        "Tutorial",
                         """
-                        Tutorial mode walks you through the analysis process
+                        The tutorial walks you through the process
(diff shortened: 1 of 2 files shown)


ekraffmiller (approved):
looks good!

PR #918, opened 2026-03-12
add errors for file size

### LINKED ISSUE #901: Profile large CSVs... and warn on upload
Right now we're loading the whole file, and at some point this will impact usability.
- Generate fake CSVs, and vary:
  - number of rows 
  - number of columns
  - bit depth / data type
- Profile performance on:
  - schema inference
  - sampling (of public datasets)
  - ...?
- Make sure none are super-linear. 
- Figure out a way to do an early warning for large files: Just read as text and count `\n`?

mccalluc:
- Fix #901

### DIFF
--- a/tests/utils/test_csv_helper.py
+++ b/tests/utils/test_csv_helper.py
@@ -109,6 +109,11 @@ def test_csv_info_none():
     assert csv_info.get_messages() == []
 
 
+def test_csv_info_no_such_file():
+    csv_info = CsvInfo(Path("no-such-file.csv"))
+    assert csv_info.get_messages() == ["No such file: no-such-file.csv"]
+
+
 def make_sparse_file(path: Path, size_in_mb: int):
     """
     >>> with NamedTemporaryFile() as tmp:
--- a/tests/utils/test_csv_helper.py
+++ b/tests/utils/test_csv_helper.py
@@ -1,7 +1,6 @@
 import csv
-import tempfile
 from pathlib import Path
-from tempfile import NamedTemporaryFile
+from tempfile import NamedTemporaryFile, TemporaryDirectory
 
 import polars as pl
 import polars.testing as pl_testing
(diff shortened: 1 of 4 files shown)


tests/utils/test_csv_helper.py
```diff
@@ -103,13 +102,56 @@ def test_csv_info(suffix, content_bytes, all, numeric, message_substring, is_err
...
+    """
+    >>> with NamedTemporaryFile() as tmp:
+    ...     path = Path(tmp.name)
+    ...     make_sparse_file(path, 1)
+    ...     print('reported size:', path.stat().st_size)
+    ...     print('blocks used:', path.stat().st_blocks)
+    reported size: 1048576
+    blocks used: 0
+    """
+    f = path.open("ab")
```
eddiestudies: with path.open("ab") as f:
    f.truncate(size_in_mb * 1024 * 1024)

eddiestudies: Generally prefer to use with blocks here. They protect against more error cases

eddiestudies: It will call the close in the end so that line could be removed

tests/utils/test_csv_helper.py
```diff
@@ -142,7 +184,7 @@ def test_datatype_inference(value_datatype):  # type: ignore
 
 
 def test_get_csv_names_mismatch():
-    with tempfile.TemporaryDirectory() as tmp:
+    with TemporaryDirectory() as tmp:
         a_path = Path(tmp) / "a.csv"
```
eddiestudies, line 187: I have written path.join so many times, this syntax is a lot better

tests/utils/test_csv_helper.py
```diff
@@ -103,13 +102,56 @@ def test_csv_info(suffix, content_bytes, all, numeric, message_substring, is_err
...
+def make_sparse_file(path: Path, size_in_mb: int):
+    """
+    >>> with NamedTemporaryFile() as tmp:
+    ...     path = Path(tmp.name)
+    ...     make_sparse_file(path, 1)
+    ...     print('reported size:', path.stat().st_size)
+    ...     print('blocks used:', path.stat().st_blocks)
+    reported size: 1048576
+    blocks used: 0
+    """
```
eddiestudies, line 126: lgtm after the with block here

eddiestudies (approved):
lgtm

Issue #922, opened 2026-04-07
Main is broken after delta=0 merge


mccalluc:
Merging 
- #897 
 
broke main. I believe its logic conflicted with
- #885 

which had been merged earlier. Minimal reproducer:
```
pytest -k 'test_make_notebook and DP_Statistics_for_2B_0'
```

(We could prevent this by requiring PRs to be up to date with main before merge, but that would be a lot more trouble for little benefit.)

<details>

```
import opendp.prelude as dp
# 30:
import polars as pl

# The OpenDP team is working to vet the core algorithms.
# Until that is complete we need to opt-in to use these features.
dp.enable_features("contrib")
# -

# + [markdown] tags=["tutorial"]
# Then define some utility functions to handle dataframes and plot results:
# -


# + tags=["tutorial"]
def round_2(number) -> float:
    """
    >>> round_2(1234)
    1200.0
    >>> round_2(0.001234)
    0.0012
    """
# 50:
    return float(f"{number:.2g}")


def make_cut_points(
    lower_bound: float, upper_bound: float, bin_count: int
) -> list[float]:
    """
    Returns one more cut point than the bin_count,
    with the cut points rounded to two decimal places
    (There are actually two more bins, extending to
# 60:
    -inf and +inf, but we'll ignore those.)

    Cut points are evenly spaced from lower_bound to upper_bound,

    >>> make_cut_points(0, 10, 2)
    [0.0, 5.0, 10.0]
    >>> make_cut_points(-1, 1, 4)
    [-1.0, -0.5, 0.0, 0.5, 1.0]

    Will return fewer points than requested if cut points
# 70:
    only differ in later decimal places.

    >>> make_cut_points(1, 1.01, 4)
    [1.0]
    """
    bin_width = (upper_bound - lower_bound) / bin_count
    # Duplicate values would cause an error in Polars.
    # Use a set to return unique values.
    return sorted({round_2(lower_bound + i * bin_width) for i in range(bin_count + 1)})


# These functions are used both in the application
# and in generated notebooks.
from polars import DataFrame


number_2b_cut_points = make_cut_points(
    lower_bound=5,
    upper_bound=15,
    bin_count=20,
)

# Use these cut points to add a new binned column to the table:
number_2b_bin_expr = (
    pl.col("2B")
# 180:
    .cut(number_2b_cut_points)  # Use "left_closed=True" to switch endpoint inclusion.
    .alias("number_2b_bin")  # Give the new column a name.
[...]

PR #923, opened 2026-04-07
delta=0 bug fix

mccalluc:
- Towards #922

If we can fix the bug and retain functionality, great, but the most important thing is to get main in working order again.


mccalluc:
Emergency bug fix so that tests pass, though this isn't necessarily the best long-term solution.

Self-merging there aren't reviewers available at this moment.

PR #927, opened 2026-04-08
simplify review workflow

### LINKED ISSUE #926: Simplify workflow
If OpenDP staff will be reviewing PRs in the future, we can simplify the workflow.

mccalluc:
If this is approved, we can also turn off "Projects" for this repo. The built-in status info on issues will be enough.

- Fix #926


mccalluc:
self-merging since we don't have outside reviewers right now.

PR #946, opened 2026-07-14
Regen screenshots and add docker to slide

mccalluc:
- Fix #920
- Only `index.md` needs review. The HTML is generated. (Could be moved to CI, but low priority.)
- Big changes to the layout should cause test failures, but smaller changes from earlier PRs didn't go over the threshold in `test_app.py`.

```
            assert (
                diff(new_path, old_path, diff_img_file=diff_path) < 0.01
            ), f"Screenshots changed! {diff_path}"
```


eddiestudies (approved):
lgtm

